PR #7, opened 2015-10-20
Make puppet wait for CSR to be signed

spiette:
This option will make puppet wait for the certificate to be signed before continuing. Else, unless the puppet master is autosigning the CSRs, the script would halt, puppet service would not be started, and packages would not be updated.


spiette:
Oops, an extra commit is in...
Closing.

PR #26, opened 2016-02-08
Several improvement

Rocco83:
Hi,

The goal of this changes are mainly remove constraint on default values for hostgroups.
- Added mydomainid and architecture_id into post_json
  - HOSTNAME and DOMAIN are now two different entities. Needed for looking up the domain.
  - get_architecture added. needed to guess OS architecture.
  - Added return_matching_domain_id. needed to guess domain ID.
  - Added return_matching_architecture_id. needed to guess OS architecture.
- bug fix
  Fixed exception in get_json (used e as variable, not declared)
- Added debug messages
  - Various function print URL if VERBOSE is set

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -23,8 +23,8 @@ def get_architecture():
   else:
      return "x86"
 
-HOSTNAME     = platform.node().split('.')[0]
-DOMAIN       = platform.node()[platform.node().index('.')+1:]
+HOSTNAME     = socket.getfqdn().split('.')[0]
+DOMAIN       = socket.getfqdn()[socket.getfqdn().index('.')+1:]
 HEXMAC       = hex(getnode())
 NOHEXMAC     = HEXMAC[2:]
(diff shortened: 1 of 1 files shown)


Rocco83:
Forgot to mention another bug fix:
-        url = opener.open(request)
  -        reply = opener.open(request)
    in post_json

bootstrap.py
```diff
@@ -15,13 +15,22 @@
...
-API_PORT  = 443
+def get_architecture():
+  # May not be safe for anything apart from 32/64 bit OS
+  is_64bit = sys.maxsize > 2**32
+  if is_64bit:
+     return "x86_64"
+  else:
+     return "x86"
+
+HOSTNAME     = platform.node().split('.')[0]
```
evgeni: please do not use `platform.node()` as it will not always return an FQDN
use `socket.getfqdn()` instead

Rocco83:
Thanks for the hint.
I have changed it to socket.getfqdn().
I have also added debug information for DOMAIN.

PR #35, opened 2016-02-22
Several fixes, mainly include --unmanaged switch

Rocco83:
Added --unmanaged parameter, managed in create_host
Added --operatingsystem as optional parameter, with function return_matching_operatingsystem_id
Fixed get_json and delete_json with "except Exception, e:" and print exception
Fixed return_matching_hg_id with verbose print myurl as other functions
Added return_matching_host for delete_json
Fixed in create_host the deletion of the previous host. The ID is needed, not FQDN.
Added in create_host the real dump of json, this one should be suitable for copy/paste to test offline with curl

In create_host the json is now dynamic for the keys.

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -45,7 +45,7 @@ parser.add_option("-a", "--activationkey", dest="activationkey", help="Activatio
 parser.add_option("-P", "--skip-puppet", dest="no_puppet", action="store_true", default=False, help="Do not install Puppet")
 parser.add_option("-g", "--hostgroup", dest="hostgroup", help="Label of the Hostgroup in Satellite that the host is to be associated with", metavar="HOSTGROUP")
 parser.add_option("-L", "--location", dest="location", default='Default_Location', help="Label of the Location in Satellite that the host is to be associated with", metavar="LOCATION")
-parser.add_option("-O", "--operatingsystem", dest="operatingsystem", default=None, help="Label of the Operating System in Satellite that the host is to be associated with", metavar="HOSTGROUP")
+parser.add_option("-O", "--operatingsystem", dest="operatingsystem", default=None, help="Label of the Operating System in Satellite that the host is to be associated with", metavar="OPERATINGSYSTEM")
 parser.add_option("-o", "--organization", dest="org", default='Default_Organization', help="Label of the Organization in Satellite that the host is to be associated with", metavar="ORG")
 parser.add_option("-S", "--subscription-manager-args", dest="smargs", default="", help="Which additional arguments shall be passed to subscription-manager", metavar="ARGS")
 parser.add_option("-u", "--update", dest="update", action="store_true", help="Fully Updates the System")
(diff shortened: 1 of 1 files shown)


bootstrap.py
```diff
@@ -45,6 +45,7 @@ def get_architecture():
 parser.add_option("-P", "--skip-puppet", dest="no_puppet", action="store_true", default=False, help="Do not install Puppet")
 parser.add_option("-g", "--hostgroup", dest="hostgroup", help="Label of the Hostgroup in Satellite that the host is to be associated with", metavar="HOSTGROUP")
 parser.add_option("-L", "--location", dest="location", default='Default_Location', help="Label of the Location in Satellite that the host is to be associated with", metavar="LOCATION")
+parser.add_option("-O", "--operatingsystem", dest="operatingsystem", default=None, help="Label of the Operating System in Satellite that the host is to be associated with", metavar="HOSTGROUP")
```
evgeni: wrong metavar

bootstrap.py
```diff
@@ -311,11 +318,23 @@ def return_matching_hg_id(hg_name):
...
     return architectureid
 
 
+def return_matching_operatingsystem_id(operatingsystem_name):
+    # Given an operatingsystem name, find its id
+    myurl = "https://" + options.sat6_fqdn + ":" + API_PORT + "/api/v2/operatingsystems?search=" + operatingsystem_name
+    if options.verbose:
+        print myurl
+    operatingsystems = get_json(myurl)
+    operatingsystemid = operatingsystems['results'][0]['id']
```
evgeni: could use some error handling if OS was not found :)

bootstrap.py
```diff
@@ -52,6 +53,7 @@ def get_architecture():
 parser.add_option("-f", "--force", dest="force", action="store_true", help="Force registration (will erase old katello and puppet certs)")
 parser.add_option("-r", "--release", dest="release", default=RELEASE, help="Specify release version")
 parser.add_option("-R", "--remove-rhn-packages", dest="removepkgs", action="store_true", help="Remove old Red Hat Network Packages")
+parser.add_option("-U", "--unmanaged", dest="unmanaged", action="store_true", help="Add the server as unmanaged. Useful to skip provisioning dependencies.")
```
evgeni: I'd not add a shortopt, IMHO shortopts should be reserved for "important" options, and having -u and -U is confusing (not that the current code would not be already guilty of that)

bootstrap.py
Rocco83: My fault, i have "backported" from a working version.

bootstrap.py
Rocco83: done :) also for architecture.

bootstrap.py
Rocco83: agreed, removed.

Rocco83:
I should have fixed all of the above, please check :)

evgeni:
thanks, LGTM :+1:

Rocco83:
I have found a small bug architecture/architectures, please wait another commit :)

bootstrap.py
```diff
@@ -375,21 +402,48 @@ def return_matching_org_label(organization):
...
+
 def create_host():
     myhgid = return_matching_hg_id(options.hostgroup)
     mylocid = return_matching_location(options.location)
     myorgid = return_matching_org(options.org)
     mydomainid = return_matching_domain_id(DOMAIN)
     architecture_id = return_matching_architecture_id(ARCHITECTURE)
-    if options.verbose:
-        print "------\nmyhgid: " + str(myhgid) + "\nmylocid: " + str(mylocid) + "\nmyorgid: " + str(myorgid) + "\nMAC: " + str(MAC) + "\nDOMAIN_ID: " + str(mydomainid) + "\nARCHITECTURE_ID: " + str(architecture_id) + "\n------"
+    host_id = return_matching_host(FQDN)
```
evgeni, line 434: I'd move that into the delete part, saves us an API call as we only need the id when deleting hosts

Rocco83, line 434: Sorry, i have not got it.
In my last revision, there is the guessing on the ID before 
`if options.force and host_id is not None:`
So the host_id have to be guessed in first place.

Better would be changing the code heavily, making in first place the check about managed/unmanaged.
If managed, architecture, operatingsystem, ... have to be retrieved.
As well, only if force = true, we sould make the other steps.
Also, i think that we should use only one function to get "ID", because the function are very similar.

But i would like to do it in the next pull request :-)

evgeni, line 434: yepp, that makes more sense now, thanks

Rocco83:
Given the above comments, i think that this pull is ready to go :)
@sideangleside do you agree?

sideangleside:
LGTM.

PR #36, opened 2016-02-24
don't hardcode ptable_id to 7

evgeni:
this was accidentally introduced in 4e1f0ad121eb27a8f09302efb07c800830b0797d

maybe @Rocco83 wants to fix it "properly", but I see no reason why the ptable should not be set on the OS/HG level.


Rocco83:
I agree on removing it. I do not even recall why this was putted in, maybe only a mistake.
I will submit a further PR for handling in the proper way.

PR #40, opened 2016-02-29
allow skipping creation of foreman hosts

evgeni:
while this is often not what one wants and --skip-puppet should be enough, sometimes one really want the bare minimum when registering the machine (esp no guessing of a proper location/hostgroup etc).


mccun934:
needs a rebase ..

evgeni:
rebased

PR #44, opened 2016-03-07
unique function to retrieve ID from satellite

Rocco83:
Fix the issue https://github.com/sideangleside/sat6-bootstrap/issues/41 .

Please doublecheck and let me know, seems working for me, but i was unable to test all conditions.

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -204,7 +204,7 @@ def clean_puppet():
 
 
 def install_puppet_agent():
-    puppet_env = return_puppetenv_for_hg(return_matching_hg_id(options.hostgroup))
+    puppet_env = return_puppetenv_for_hg(return_matching_id('hostgroups', 'title=%s' % options.hostgroup, False))
     print_generic("Installing the Puppet Agent")
     exec_failexit("/usr/bin/yum -y install puppet")
     exec_failexit("/sbin/chkconfig puppet on")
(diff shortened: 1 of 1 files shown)


bootstrap.py
```diff
@@ -300,61 +302,24 @@ def delete_json(url):
...
         sys.exit(2)
 
-
-def return_matching_domain_id(domain_name):
-    # Given a domain, find its id
-    myurl = "https://" + options.sat6_fqdn + ":" + API_PORT + "/api/v2/domains/?" + urlencode([('search', 'name=%s' % domain_name)])
+# Search in API 
+# given a search key, return the ID
+# api_name is the key in url for API name, search_key must contain also the key for search (name=, title=, ...)
+def return_matching_id(api_name, search_key, null_result_ok):
```
evgeni: maybe make `null_result_ok=False` the default?

bootstrap.py
```diff
@@ -368,7 +333,6 @@ def return_puppetenv_for_hg(hg_id):
     else:
         return 'production'
 
-
```
evgeni, line 371: pep8 requires two empty lines between functions

evgeni:
I think you missed a few occasions:
- https://github.com/Rocco83/sat6-bootstrap/blob/master/bootstrap.py#L336
- https://github.com/Rocco83/sat6-bootstrap/blob/master/bootstrap.py#L370
- https://github.com/Rocco83/sat6-bootstrap/blob/master/bootstrap.py#L438

otherwise LGTM :+1:

Rocco83:
Thanks @evgeni for pointing out.
New code, not yet tested, i will do it ASAP... if someone else is willing to, very welcome :)

Rocco83:
Forgot to add null_result_ok=False previously. the previous comment apply anyway.

Rocco83:
I made a couple of test, seems working without problem.

PR #49, opened 2016-03-18
Quote fix

rplevka:
- quoted the location and organization values in the url to support multi-word values, like "Default Organization", "Default Location", etc.

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -346,7 +346,7 @@ def return_matching_org(organization):
 
 def return_matching_org_label(organization):
     # Given an org name, find its label - required by subscription-manager
-    myurl = "https://" + options.foreman_fqdn + ":" + API_PORT + "/katello/api/organizations/" + organization
+    myurl = "https://" + options.foreman_fqdn + ":" + API_PORT + "/katello/api/organizations/" + urllib2.quote(organization)
     if options.verbose:
         print "myurl: " + myurl
     organization = get_json(myurl)


Rocco83:
Hi,

i think that this will be superseeded by 
https://github.com/sideangleside/sat6-bootstrap/pull/44

Do you agree?

rplevka:
@Rocco83, sure, as long as it fixes the multi-word string problem. Go ahead and close this PR.

Rocco83:
I cannot, unfortunately :( 
@rplevka can you do it on your own?
@sideangleside are you still able to do it?

evgeni:
imho this was not fixed with @Rocco83's stuff in #44 

@rplevka can you retry and maybe rebase your fix please?

mccun934:
needs a rebase ..

rplevka:
ok, rebased and modified my commit

Rocco83:
@rplevka as we did for https://github.com/Katello/client-bootstrap/pull/44 maybe we can create an even more generic function that cover search for name and labels, adding the key of return json to be returned.

Replacing return_matching_key() in https://github.com/rplevka/sat6-bootstrap/blob/quote_fix/bootstrap.py#L303 with something like:
`return_matching_key(api_name, search_key, null_result_ok=False, return_key="id"):`

mrmattx3:
In relation to #52 ... I made the change here:

myurl = "https://" + options.foreman_fqdn + ":" + API_PORT + "/katello/api/organizations/" + urllib2.quote(organization)

Looks like the same return...

[bootstrap_log_verbose_v2.txt](https://github.com/Katello/client-bootstrap/files/200307/bootstrap_log_verbose_v2.txt)

Rocco83:
Please try to change in https://github.com/Katello/client-bootstrap/blob/master/bootstrap.py#L360
`myorgid = return_matching_id('organizations', 'name=%s' % options.org, False)`
in
`myorgid = return_matching_id('organizations', 'name="%s"' % options.org, False)`
see double quote after `name=`, before and after `%s`.

mrmattx3:
Good news!  that did fix the issue at hand (with the space).  I actually seem to be having some issue with my host groups now, and that's on me to fix...

[RUNNING], [2016-04-01 15:40:35], [Calling Foreman API to create a host entry associated with the group, org & location]
Error: cannot connect to the API: HTTP Error 422: Unprocessable Entity
Check your URL & try to login using the same user/pass via the WebUI and check the error!
error: {
  "error": {"id":null,"errors":{"operatingsystem_id":["can't be blank"],"root_pass":["should be 8 characters or more","should not be blank - consider setting a global or host group default"],"ptable_id":["can't be blank unless a custom partition has been defined"]},"full_messages":["Operatingsystem can't be blank","Root password should be 8 characters or more","Root password should not be blank - consider setting a global or host group default","Partition Table can't be blank unless a custom partition has been defined"]}

Thanks so much for the help!

Rocco83:
means that every like calling `return_matching_id()` must have double quote especially if the search key is 'name'.
for 'label', seems to me that satellite automatically replace `" "` with `"_"`
@rplevka do you think you can handle all of the previous topic in this PR?

rplevka:
@Rocco83 honestly, perhaps we're getting to a point where we could use [nailgun](http://nailgun.readthedocs.org/en/latest/#quick-start)
straight away

Rocco83:
@rplevka the use case for bootstrap.py imo is a client with a minimal footprint to be registered in satellite.
having too much code (nailgun) would not respect this use case.
but imo this is not the right place, if you think so maybe the best place is a new issue.

evgeni:
@rocco83 is right. The code is supposed to work on a vanilla rhel box, without the need to install anything as install might need a sub which we are just about to add. Chickens and eggs all the way down.

mccun934:
ACK, looks good. thanks @evgeni

PR #51, opened 2016-03-31
fixes #50 - replaced deprecated yum localinstall by rpm command

### LINKED ISSUE #50: get_bootstrap_rpm doesn't die in case if remote package doesn't exist
`get_bootstrap_rpm` calls `exec_failexit` on `yum localinstall` but due to way `yum` handles responses, it does not catch the failure in download of the package:

``` bash
> yum -y localinstall http://completely.fictive.hostname/foo.rpm --nogpgcheck
Loaded plugins: product-id, security, subscription-manager
Setting up Local Package Process
Cannot open: http://completely.fictive.hostname/foo.rpm. Skipping.
Nothing to do
> echo $?
0
```
- so it evaluates this situation as a success and continues.

rplevka:
Addresses #50 .

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -148,7 +148,7 @@ def get_bootstrap_rpm():
     if options.force:
         clean_katello_agent()
     print_generic("Retrieving Client CA Certificate RPMs")
-    exec_failexit("/usr/bin/yum -y localinstall http://%s/pub/katello-ca-consumer-latest.noarch.rpm --nogpgcheck" % options.foreman_fqdn)
+    exec_failexit("rpm -Uvh http://%s/pub/katello-ca-consumer-latest.noarch.rpm" % options.foreman_fqdn)
 
 
 def migrate_systems(org_name, activationkey):
@@ -178,6 +178,7 @@ def register_systems(org_name, activationkey, release):
 def unregister_system():
     print_generic("Unregistering")
     exec_failexit("/usr/sbin/subscription-manager unregister")
+
 
 def clean_katello_agent():
     print_generic("Removing old Katello agent and certs")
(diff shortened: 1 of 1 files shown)


evgeni:
IMHO this should be `rpm -ivh http://%s/pub/katello-ca-consumer-latest.noarch.rpm` to match what we do in kickstart and support RHEL5 which does not have this functionality.

evgeni:
ah, one thing:
drop the path please, it's /bin/rpm on rhel5/6 but /usr/bin/rpm on rhel7 (due to merged /bin:/usr/bin)

evgeni:
:+1:

CalvinHartwell:
> IMHO this should be rpm -ivh http://%s/pub/katello-ca-consumer-latest.noarch.rpm to match what we do in kickstart and support RHEL5 which does not have this functionality.

Should also limit any future issues with DNF? :+1:

mccun934:
needs a rebase

rplevka:
@mccun934 rebased

Rocco83:
@evgeni based on http://www.linuxquestions.org/questions/linux-newbie-8/rpm-uvh-or-ivh-116706/ why should we use ivh instead of uvh?

evgeni:
@Rocco83 mainly to match what we do in Kickstart:
https://github.com/Katello/katello/blob/master/app/views/foreman/unattended/snippets/_subscription_manager_registration.erb

Rocco83:
interesting. my question btw is the same, but apply also to kickstart :)
  rpm -ivh <%= subscription_manager_configuration_url(@host) %>
why -ivh and not -Uvh?

rplevka:
@Rocco83 After browsing some official documentation and knowledge bases I noticed they also use `-Uvh` so I decided to go with your suggestion.
- also fixed few (unrelated) formatting nitpicks my linter has highlighted

evgeni:
:+1:

LGTM, ship it :)

evgeni:
oh, wait! it has sat6_fqdn again, that's foreman_fqdn now!

rplevka:
@evgeni, d'oh...good catch. let me fix that real quick

Done!

rplevka:
@evgeni done

evgeni:
:)
you also probably want to look at the print line for "Retrieving Candlepin Consumer RPMs", I think @mccun934 changed it on purose

evgeni:
good now :)

rplevka:
@evgeni, ok, done - undone that change - can you review once again, pls? :)

Rocco83:
seems good to me as well :)

PR #56, opened 2016-04-07
clean up environment before doing anything

evgeni:
this should prevent $tools to load broken libssl etc

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -197,6 +197,11 @@ def clean_puppet():
     exec_failexit("rm -rf /var/lib/puppet/")
 
 
+def clean_environment():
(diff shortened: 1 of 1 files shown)


bootstrap.py
```diff
@@ -197,6 +197,12 @@ def clean_puppet():
     exec_failexit("rm -rf /var/lib/puppet/")
 
 
+def clean_environment():
+    for key in ['LD_LIBRARY_PATH', 'LD_PRELOAD']:
+      if key in os.environ:
+        os.environ.pop(key)
```
rplevka:
``` python
if key in os.environ:
        os.environ.pop(key)
```

can be simply written as

``` python
os.environ.pop(key, None)
```

ref: https://docs.python.org/2/library/stdtypes.html#dict.pop

PR #58, opened 2016-04-11
stricter check if machine is subscribed to legacy management

evgeni:
in some cases the machine will have a systemid file but that will be outdated/revoked/whatever, which will result in failures during migration

Signed-off-by: Evgeni Golov evgeni@golov.de

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -397,7 +397,11 @@ def disassociate_host(host_id):
 
 
 def check_rhn_registration():
-    return os.path.exists('/etc/sysconfig/rhn/systemid')
+    if os.path.exists('/etc/sysconfig/rhn/systemid'):
+        retcode = commands.getstatusoutput('rhn-channel -l')[0]
+        return retcode == 0
+    else:
+        return False
 
 
 def get_api_port():


bootstrap.py
```diff
@@ -397,7 +397,12 @@ def disassociate_host(host_id):
 
 
 def check_rhn_registration():
-    return os.path.exists('/etc/sysconfig/rhn/systemid')
+    if os.path.exists('/etc/sysconfig/rhn/systemid'):
+        output = commands.getstatusoutput('rhn-channel -l')
+        retcode = output[0]
+        return retcode == 0
```
rplevka, line 402: can't we simply put that on a single line?

``` python
return commands.getstatusoutput('rhn-channel -l')[0] == 0
```

Just asking, not sure what's our priority. Not a blocker

rplevka:
ACK

bootstrap.py
evgeni, line 402: syntax wise: sure. I'd prefer the more readable way. maybe

``` python
retcode =  commands.getstatusoutput('rhn-channel -l')[0]
return retcode == 0
```

?

rplevka, line 402: Either way is good to me really.

Issue #57, opened 2016-04-09
error in return_matching_id print_error


Rocco83:
The line
https://github.com/Katello/katello-client-bootstrap/blob/master/bootstrap.py#L317
have a syntax error

Fixed in https://github.com/Rocco83/sat6-bootstrap/blob/master/bootstrap.py#L318
PR will follow

Issue #55, opened 2016-04-07
Make usage of a more generic function


Rocco83:
Based on https://github.com/Katello/client-bootstrap/pull/44 and https://github.com/Katello/client-bootstrap/pull/49#issuecomment-204557946 
the current function
`return_matching_id(api_name, search_key, null_result_ok=False):`
should became something like
`return_matching_key(api_name, search_key, null_result_ok=False, return_key="id"):`

Following functions should became useless:
https://github.com/Katello/client-bootstrap/blob/master/bootstrap.py#L321
https://github.com/Katello/client-bootstrap/blob/master/bootstrap.py#L333
https://github.com/Katello/client-bootstrap/blob/master/bootstrap.py#L347

Anyone willing to check if there anything else impacted is more than welcome :)

Rocco83:
Some proof of concept in https://github.com/Rocco83/sat6-bootstrap/commit/ceb8f6a8130f14481ff3698b55a95d1f0462b31c this is not currently doing nothing with return_key.

Rocco83:
I came up in one issue.
-  subscription-manager want label for organizations
-  label is not present in the return values of Foreman API, but are present in Katello API.
  Reference: Satellite 6.1.7

I see three path possible:
1. adding the API needed in every call
`return_matching_key(api_name, api_kind, search_key, null_result_ok=False, return_key="id"):` 
2. create one fixed array of keys, when the api_name will be (in example) organization, it will point to katello
3. instead of api_name, we ask for api_path (ie: `/katello/api/organizations`)

Do you see any other solution?

I personally prefer the third.
Anyone? @evgeni @mccun934 @rplevka ?
Any comment is more than welcome.

Example:
https://fqdn/api/v2/organizations/

```
{
  "total": 3,
  "subtotal": 3,
  "page": 1,
  "per_page": 20,
  "search": null,
  "sort": {
    "by": null,
    "order": null
  },
  "results": [{"id":1,"name":"name","title":"title","created_at":"2015-12-09T11:49:06Z","updated_at":"2016-02-10T10:32:01Z"}]
}
```

https://fqdn/katello/api/organizations

```
{
  "total": 3,
  "subtotal": 3,
  "page": 1,
  "per_page": 20,
  "search": null,
  "sort": {
    "by": null,
    "order": null
  },
[...]

Issue #54, opened 2016-04-07
Fix quoting in return_matching_id()


Rocco83:
Based on https://github.com/Katello/client-bootstrap/pull/49#issuecomment-204562314 quoting must be fixed for every call to return_matching_id().
More information https://github.com/Katello/client-bootstrap/pull/49#issuecomment-204559470

Rocco83:
This should be fixed in https://github.com/Rocco83/sat6-bootstrap/commit/ceb8f6a8130f14481ff3698b55a95d1f0462b31c
PR will follow

Issue #53, opened 2016-04-05
loose the hardcoded error message as it produces confusing errors


rplevka:
Perhaps the following lines of code should be dropped:
https://github.com/Katello/client-bootstrap/blob/master/bootstrap.py#L241-#L242
as they provide confusing output if the root cause is completely different from 'connecting to API':

e.g.:

``` bash
[RUNNING], [2016-04-05 14:29:38], [Calling Satellite API to create a host entry associated with the group, org & location] 
Error: cannot connect to the API: HTTP Error 422: Unprocessable Entity
Check your URL & try to login using the same user/pass via the WebUI and check the error!
error: {
  "error": {"id":null,"errors":{"name":["has already been taken"],"mac":["has already been taken"]},"full_messages":["Name has already been taken","MAC address has already been taken"]}
}
```

Rocco83:
What about handling the HTTP error 401 vs 422 vs others?

rplevka:
Sure, that would address the issue as well

rplevka:
@Rocco83 anyway, I think dropping the 'cannot connect to the API' message would do (at least for now). The other errors seem to be self explaining.
What do you think?

Rocco83:
@rplevka :+1: please check https://github.com/Katello/katello-client-bootstrap/issues/61

Rocco83:
@rplevka can you check and close the issue if fixed?

evgeni:
also #67 should help here :)

rplevka:
This has been fixed already.

PR #61, opened 2016-04-26
Generic function to call API, quote fix, 

Rocco83:
- fixed quoting in all API call functions
- fixed typos
- created wrapper for two api foreman and katello (like that to keep
  backward compatibility)
- removed return_matching_org_label
- fixed message in default error for API
- added messages for HTTP error 401 and 422

This should address and fix:
https://github.com/Katello/katello-client-bootstrap/issues/57
https://github.com/Katello/katello-client-bootstrap/issues/55
https://github.com/Katello/katello-client-bootstrap/issues/54
https://github.com/Katello/katello-client-bootstrap/issues/53

and superseed:
https://github.com/Katello/katello-client-bootstrap/pull/59
https://github.com/Katello/katello-client-bootstrap/pull/60

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -152,7 +152,8 @@ def get_bootstrap_rpm():
 
 
 def migrate_systems(org_name, activationkey):
-    org_label = return_matching_org_label(org_name)
+    #org_label = return_matching_org_label(org_name)
+    org_label = return_matching_foreman_key('organizations', 'name="%s"' % org_name, "label", False)
     print_generic("Calling rhn-migrate-classic-to-rhsm")
     options.rhsmargs += " --destination-url=https://%s:%s" % (options.foreman_fqdn, API_PORT)
     if options.legacy_purge:
@@ -166,7 +167,8 @@ def migrate_systems(org_name, activationkey):
 
 
 def register_systems(org_name, activationkey, release):
-    org_label = return_matching_org_label(org_name)
+    #org_id = return_matching_org_label(org_name)
+    org_label = return_matching_katello_key('organizations', 'name="%s"' % org_name, "label", False)
     print_generic("Calling subscription-manager")
     options.smargs += " --serverurl=https://%s:%s/rhsm --baseurl=https://%s/pulp/repos" % (options.foreman_fqdn, API_PORT, options.foreman_fqdn)
     if options.force:
@@ -198,7 +200,7 @@ def clean_puppet():
 
 
 def install_puppet_agent():
-    puppet_env = return_puppetenv_for_hg(return_matching_id('hostgroups', 'title=%s' % options.hostgroup, False))
+    puppet_env = return_puppetenv_for_hg(return_matching_foreman_key('hostgroups', 'title="%s"' % options.hostgroup, 'id', False))
     print_generic("Installing the Puppet Agent")
     exec_failexit("/usr/bin/yum -y install puppet")
     exec_failexit("/sbin/chkconfig puppet on")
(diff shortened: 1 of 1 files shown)


Rocco83:
@rplevka please test it in your environment and check if this is ok
@evgeni do you think is better to push this commit after #59 and #60 ?
Some code cleanup is for sure needed, but the code is actually working in my environment.

evgeni:
@Rocco83 technically the code in #59 and #60 is identical to yours. so it wouldn't matter. I just opened these to get the fixes for the actual bugs quicker, but that did not happen anyways.

bootstrap.py
```diff
@@ -152,7 +152,8 @@ def get_bootstrap_rpm():
 
 
 def migrate_systems(org_name, activationkey):
-    org_label = return_matching_org_label(org_name)
+    #org_label = return_matching_org_label(org_name)
+    org_label = return_matching_foreman_key('organizations', 'name="%s"' % org_name, "label", False)
```
evgeni: that surely should be `return_matching_katello_key`, no?

evgeni: I'd prefer `'label'` instead of `"label"` to be consistent with other code

bootstrap.py
```diff
@@ -166,7 +167,8 @@ def migrate_systems(org_name, activationkey):
 
 
 def register_systems(org_name, activationkey, release):
-    org_label = return_matching_org_label(org_name)
+    #org_id = return_matching_org_label(org_name)
+    org_label = return_matching_katello_key('organizations', 'name="%s"' % org_name, "label", False)
```
evgeni: I'd prefer `'label'` instead of `"label"` to be consistent with other code

evgeni:
the three commented `return_matching_org_label` can be deleted, right?

bootstrap.py
```diff
@@ -237,8 +239,18 @@ def get_json(url):
         request.add_header("Authorization", "Basic %s" % base64string)
         result = urllib2.urlopen(request)
         return json.load(result)
+    except urllib2.HTTPError, e:
+        if e.code == 401:
```
evgeni, line 246: I think this will eat all HTTPErrors that are not 401/422 and never display them. That sounds wrong.
Also it should say "Foreman", not "satellite" ;)

bootstrap.py
Rocco83: Done both, thanks for the hint.

bootstrap.py
Rocco83, line 246: Added a raise. I have not tested the code, but should work.
What do you think?
My thought is that the full try catch need to be rewritten, but this can wait for now.

bootstrap.py
Rocco83: done.

Rocco83:
About return_matching_org_label, that was exactly the meaning of "Some code cleanup is for sure needed" :)

evgeni:
@Rocco83 please rebase :)

Rocco83:
Done!

bootstrap.py
```diff
@@ -268,8 +278,19 @@ def post_json(url, jdata):
         request.add_header("Accept", "application/json")
         request.get_method = lambda: 'POST'
         reply = opener.open(request)
+    except urllib2.HTTPError, e:
+        if e.code == 401:
```
rplevka, line 282: I would drop all these hooks and simply raised one general HTTPError with printing out the `e.read` as the api itself contains the error message, e.g.:

```
error: {
  "error": {"message":"Unable to authenticate user admin"}
}
```
- I think it is self explaining and does not require you to handle all sorts of return codes. Or am I missing something?

Rocco83, line 282: We may have the following (just a 2 minute brainstorming):
1) wrong request made (err 400, POST-GET(?)-DELETE)
2) wrong username/password (err 403, POST-GET-DELETE hardcoding "admin" is just wrong)
3) http refused connection (connection reset, code?)
4) firewall (connection timeout, code?)

and so on... am i wrong?

rplevka, line 282: 422 (unprocessable entity),
500 (ISE),
502,503,504 - bad gateway, service unavail., gateway timeout,...
even this one:  418 - I'm a teapot (not likely, but who knows? :D )
etc.

evgeni, line 282: let's move the discussion to #62 :)

Issue #65, opened 2016-05-06
RHEL5 support


Rocco83:
Hi,

I am testing finally the script on some RHEL5, python 2.4.

Got, up to now, the following problems:
- json python non existing.
  Solution: http://stackoverflow.com/questions/3291682/json-module-for-python-2-4
  Status: Tested, working
- sys.maxsize non existing.
  Solution: http://stackoverflow.com/questions/1405913/how-do-i-determine-if-my-python-shell-is-executing-in-32bit-or-64bit-mode-on-os
  Status: testes only on command line, seems working
- platform.linuxdistribution non existing
  Solution: https://docs.python.org/2/library/platform.html#platform.linux_distribution we can use platform.dist on old systems
  Status: Briefly tested, working

In order to get the version and understand which code to be run, in example for platform.linuxdistribution,
https://docs.python.org/2/library/sys.html#sys.hexversion
or
http://stackoverflow.com/questions/1093322/how-do-i-check-what-version-of-python-is-running-my-script

patch coming soon, this has been not tested deeply.

evgeni:
I have a long standing branch on this: https://github.com/evgeni/sat6-bootstrap/tree/rhel5
never merged, though

Rocco83:
Well, shame on me... i have not checked the fork :)
Discussing about the code:
- 64 bit check

yours https://github.com/Katello/katello-client-bootstrap/compare/master...evgeni:rhel5?expand=1#diff-cd48c161c09e850299d4a00383422690R25
mine https://github.com/Katello/katello-client-bootstrap/pull/66/commits/e0469f3eadd16a3813db96738898fad736ae465d#diff-cd48c161c09e850299d4a00383422690R25
Your approach is cleaner, mine is more general (but maybe not needed)
- platform.linux_distribution

yours https://github.com/Katello/katello-client-bootstrap/compare/master...evgeni:rhel5?expand=1#diff-cd48c161c09e850299d4a00383422690R56
mine https://github.com/Katello/katello-client-bootstrap/pull/66/commits/e0469f3eadd16a3813db96738898fad736ae465d#diff-cd48c161c09e850299d4a00383422690R48
Your approach do not take care of the version, which can be good. mine is stricter as approach, and may lead to some bugs.
- json

your https://github.com/Katello/katello-client-bootstrap/compare/master...evgeni:rhel5?expand=1#diff-cd48c161c09e850299d4a00383422690R513
[...]

PR #66, opened 2016-05-06
Fixed to support python 2.4

Rocco83:
Hi,

as by https://github.com/Katello/katello-client-bootstrap/issues/65


Rocco83:
Based on https://github.com/Katello/katello-client-bootstrap/issues/65 is better to use the @evgeni code :)

PR #68, opened 2016-05-09
RHEL5 support

### LINKED ISSUE #65: RHEL5 support

evgeni:
this adds support for RHEL5 which is a bit challenging:
- Python 2.4 has no `json` module, thus we try to use `python-simplejson` (and even try to `yum install` it, if it is not available)
- Python 2.4 has no `uuid` module, thus try to get the MAC address from `/sys/class/net` in that case
- Python 2.4 has no `platform.linux_distribution()`, thus try `platform.dist()` in that case. This is not as reliable, but should work.
- Python 2.4 has no `sys.maxsize`, add a workaround.
- Puppet on RHEL5 has no `puppet config set`, so write `puppet.conf` using a simple template instead.
- subscription-manager migration code on RHEL5 is a bit old and outdated, thus "migrate" machines by doing a clean new registration in that case

closes #65

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -1,6 +1,9 @@
 #!/usr/bin/python
 #
-import json
+try:
+    import json
+except ImportError:
+    import simplejson as json
 import getpass
 import urllib2
 import base64
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -21,7 +21,10 @@ from ConfigParser import SafeConfigParser
 
 def get_architecture():
     # May not be safe for anything apart from 32/64 bit OS
-    is_64bit = sys.maxsize > 2 ** 32
+    try:
+        is_64bit = sys.maxsize > 2 ** 32
+    except:
+        is_64bit = platform.architecture()[0] == '64bit'
     if is_64bit:
         return "x86_64"
     else:
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -1,9 +1,6 @@
 #!/usr/bin/python
 #
-try:
-    import json
-except ImportError:
-    import simplejson as json
+
 import getpass
 import urllib2
 import base64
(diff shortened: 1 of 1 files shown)


Rocco83:
LGTM, as previously discussed :+1:

peppos:
Hi, I try to test the script in RHEL5 and I have this errror:

  File "bootstrap_rhel5.py", line 531, in ?
    prepare_rhel5_migration()
  File "/bootstrap_rhel5.py", line 486, in prepare_rhel5_migration
    me.options.force = True
AttributeError: 'MigrationEngine' object has no attribute 'options'

evgeni:
@peppos thanks for testing, I will have a look at that shortly.

evgeni:
@peppos can you try the new version please? :)

peppos:
Hi, thanks @evgeni. I try it and I have another error.

Problem encountered getting the list of subscribed channels.  Exiting.

My system is not registered. I think the problem is given by

> subscribed_channels = me.get_subscribed_channels_list()

in line 490

evgeni:
@peppos okay, that's a case we never tried I think. So you have no `/etc/sysconfig/rhn/systemid` and `rhn-channel -l` will return an error, right?

peppos:
Yes @evgeni. So, because in my environment the server is not registered to rhn I try this workarround:
change 

> if not options.remove and int(RELEASE[0]) == 5:

to

> if not options.remove and int(RELEASE[0]) == 4:

for bypass the migration and it work fine. Is possible to introduce a check?

```
if server is registered --> migrate it
else --> not migrate
```

evgeni:
@peppos Did you have any content in `/etc/pki/product/`? The problem is that old RHEL5 installs have no `/etc/pki/product/69.pem` by default, and then `subscription-manager` will fail terribly. So that step you skipped is actually needed for both, migrations and new-regs. At least in some cases :)

peppos:
@evgeni yes, I have a `69.pem` file in`/etc/pki/product/`. I will try on another server.

evgeni:
@peppos ok, I think 1aa8e2c6234893a447807a59a7e2fad4e611402a should fix your case, mind giving it a go? :)

peppos:
Thanks @evgeni

PR #69, opened 2016-05-12
dont fail if we do not know the domain, maybe the HG can help

evgeni:
but expect the user to pass --force or --skip-puppet, as not having a working FQDN can lead to funny results


evgeni:
rebased after the latest changes

PR #75, opened 2016-05-17
help should say we are looking for name/title, not label

evgeni:
these are three different things and that might confuse the user otherwise.


mccun934:
FYI, for future PRs we are going to work towards following Katello's standard of squashing all commits in a PR down to 1 but I won't require it on this repo yet.

bootstrap.py
```diff
@@ -72,11 +72,11 @@ def get_architecture():
...
-parser.add_option("-g", "--hostgroup", dest="hostgroup", help="Label of the Hostgroup in Foreman that the host is to be associated with", metavar="HOSTGROUP")
-parser.add_option("-L", "--location", dest="location", default='Default_Location', help="Label of the Location in Foreman that the host is to be associated with", metavar="LOCATION")
-parser.add_option("-O", "--operatingsystem", dest="operatingsystem", default=None, help="Label of the Operating System in Foreman that the host is to be associated with", metavar="OPERATINGSYSTEM")
-parser.add_option("--partitiontable", dest="partitiontable", default=None, help="Label of the Operating System in Foreman that the host is to be associated with", metavar="PARTITIONTABLE")
-parser.add_option("-o", "--organization", dest="org", default='Default_Organization', help="Label of the Organization in Foreman that the host is to be associated with", metavar="ORG")
+parser.add_option("-g", "--hostgroup", dest="hostgroup", help="Title of the Hostgroup in Foreman that the host is to be associated with", metavar="HOSTGROUP")
```
mccun934, line 75: there is no "Title" field associated to Hostgroup, why isn't this help text referring to "Name"?

hg-name

bootstrap.py
```diff
@@ -72,11 +72,11 @@ def get_architecture():
...
+parser.add_option("-L", "--location", dest="location", default='Default Location', help="Title of the Location in Foreman that the host is to be associated with", metavar="LOCATION")
+parser.add_option("-O", "--operatingsystem", dest="operatingsystem", default=None, help="Title of the Operating System in Foreman that the host is to be associated with", metavar="OPERATINGSYSTEM")
+parser.add_option("--partitiontable", dest="partitiontable", default=None, help="Name of the Operating System in Foreman that the host is to be associated with", metavar="PARTITIONTABLE")
```
mccun934, line 78: same thing for OS and Location, shouldn't this be Name vs Title?

bootstrap.py
evgeni, line 75: When you talk to the API there is both, name and title. Name is whatever you enter in the form you posted above, title is the name, including partents.
So e.g. you have a HG called "test" and it's parent is "example", name will be "test" and title will be "example/test". We need the title here, as we potentially want to deploy into nested HGs.

bootstrap.py
evgeni, line 78: same answer, those can be nested.

mccun934, line 78: understood, appreciate the clarification

mccun934:
ACK - merging

PR #80, opened 2016-06-01
Fix disassociate host

evgeni:
- `put_json` was not defined, so it could not be called
- `subscription-manager` might error out if the host was deleted before unregistering

thanks to @bergsjoh for spotting the issue and providing the patch

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -331,6 +331,9 @@ def delete_json(url):
     return call_api(url, method='DELETE')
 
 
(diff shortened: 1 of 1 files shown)


lzap:
Oh this broke Satellite 6.2 latest snap:

https://bugzilla.redhat.com/show_bug.cgi?id=1342040

lzap:
I mean, it still does not work for me.

PR #81, opened 2016-06-02
Multiple architectures, error reporting and 201 fix

lzap:
I am doing multiple architectures testing via bootstrap script and I hit several issues I am going to fix in this PR.

1) Architecture was fixed only to 64 and 32 bit intels, also the 32 variant was not named right. This now uses uname to deliver correct architecture names as reported by Puppet / Facter.

2) Download only for yum. I am testing this script with lazy repositories with multiple concurrent connections and there seems to be some limitations in Satellite 6.2 beta which makes Pulp/Squid to fail with 403 or 404 errors until the RPM package is properly loaded. New option allows to call yum with `--downloadonly` multiple times until it returns correctly.

3) Satellite returns 201 when host already exists during creation call, but the Python built-in library treats 201 as an error which is incorrect. New provided which handles this correctly is supplied to prevent this kind of mis-behavior.

I am still currently testing this patch, but feel free to review now.

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -1,5 +1,4 @@
 #!/usr/bin/python
-#
 
 import getpass
 import urllib2
@@ -16,17 +15,8 @@ from optparse import OptionParser
 from urllib import urlencode
 from ConfigParser import SafeConfigParser
 
-
 def get_architecture():
-    # May not be safe for anything apart from 32/64 bit OS
-    try:
-        is_64bit = sys.maxsize > 2 ** 32
-    except:
-        is_64bit = platform.architecture()[0] == '64bit'
-    if is_64bit:
-        return "x86_64"
-    else:
-        return "x86"
+    return os.uname()[4]
 
 FQDN = socket.getfqdn()
 if FQDN.find(".") != -1:
@@ -161,6 +151,7 @@ def exec_failok(command):
         print_warning(command)
     print output[1]
     print ""
+    return retcode
 
 
 def exec_failexit(command):
@@ -175,12 +166,14 @@ def exec_failexit(command):
     print_success(command)
     print ""
 
+def yum(command, pkgs = ""):
+    exec_failexit("/usr/bin/yum -y %s %s" % (command, pkgs))
 
 def install_prereqs():
     print_generic("Installing subscription manager prerequisites")
-    exec_failexit("/usr/bin/yum -y remove subscription-manager-gnome")
-    exec_failexit("/usr/bin/yum -y install subscription-manager subscription-manager-migration-*")
-    exec_failexit("/usr/bin/yum -y update yum openssl")
+    yum("remove", "subscription-manager-gnome")
+    yum("install", "subscription-manager subscription-manager-migration-*")
+    yum("update", "yum openssl")
 
 
 def get_bootstrap_rpm():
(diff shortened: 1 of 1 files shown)


bootstrap.py
```diff
@@ -11,22 +10,16 @@
...
-    # May not be safe for anything apart from 32/64 bit OS
-    try:
-        is_64bit = sys.maxsize > 2 ** 32
-    except:
-        is_64bit = platform.architecture()[0] == '64bit'
-    if is_64bit:
-        return "x86_64"
-    else:
-        return "x86"
+    p = subprocess.Popen(["uname", "-m"], stdout=subprocess.PIPE)
```
evgeni: why not `os.uname()[4]`?

lzap: > why not `os.uname()[4]`?

Sure, thanks.

## 

Later,
 Lukas #lzap Zapletal

bootstrap.py
```diff
@@ -465,8 +475,9 @@ class MEOptions:
     # doublecheck and copy it manually if not
     if not os.path.exists('/etc/pki/product/'):
         os.mkdir("/etc/pki/product/")
-    if not os.path.exists('/etc/pki/product/69.pem'):
-        for line in open("/usr/share/rhsm/product/RHEL-5/channel-cert-mapping.txt"):
+    mapping_file = "/usr/share/rhsm/product/RHEL-5/channel-cert-mapping.txt"
+    if not os.path.exists('/etc/pki/product/69.pem') and os.path.exists(mapping_file):
+        for line in open(mapping_file):
             if ("rhel-x86_64-server-5:" in line and ARCHITECTURE == "x86_64") or ("rhel-i386-server-5:" in line and ARCHITECTURE == "x86"):
```
evgeni: as you fixed the content of `ARCHITECTURE`, I think this should be something like:

```
if line.startswith('rhel-%s-server-5' % ARCHITECTURE):
…
```

bootstrap.py
```diff
@@ -174,12 +167,18 @@ def exec_failexit(command):
     print_success(command)
     print ""
 
+def yum(command, pkgs = ""):
+    for x in range(0, int(options.download)):
+        if exec_failok("/usr/bin/yum -y --downloadonly %s %s" % (command, pkgs)):
```
evgeni: `--downloadonly` is not always available (it was provided by an optional plugin in older yum versions)

bootstrap.py
lzap: Fixed.

bootstrap.py
lzap: Yeah, I know. But since this is optional flag which is turned of by default, can we assume user knows this (otherwise yum will print an error which will be ignored anyway).

bootstrap.py
```diff
@@ -209,7 +208,8 @@ def register_systems(org_name, activationkey, release):
     options.smargs += " --serverurl=https://%s:%s/rhsm --baseurl=https://%s/pulp/repos" % (options.foreman_fqdn, API_PORT, options.foreman_fqdn)
     if options.force:
         options.smargs += " --force"
-    exec_failexit("/usr/sbin/subscription-manager register --org '%s' --name '%s' --activationkey '%s' %s" % (org_label, FQDN, activationkey, options.smargs))
+    # RHEL 5 sometimes register successfully with "Error updating system data on the server" error
```
evgeni: can we make that rhel5 only? (or fix sub-man in rhel5? :))
I think allowing sub-man to fail randomly and ignoring it is a bad idea for a tool that is supposed to subscribe the machine

evgeni:
Shouldn't HTTP 201 be returned when the request _triggered_ the creation, and if the host already existed it should be rather something in the 4xx or 5xx range?

lzap:
Sorry 201 is actually just "host created". I don't know why it did not work for me, looks like some Python version was not compatible.

I don't know when to get back to this, what exactly do you want to separate into another PR?

evgeni:
I think the `get_architecture()` and `mapping_file` changes are straiht-forward to apply. 

The others idk, @sideangleside @mccun934 say something :)

lzap:
Ok I removed download parameter and only keeping architectures, better error reporting and 201 issue (http://stackoverflow.com/questions/7032890/why-does-pythons-urllib2-urlopen-raise-an-httperror-for-successful-status-cod/7033056) in there.

lzap:
Resolved conflicts, I am also keeping the "yum" helper function. Feels better.

PR #85, opened 2016-06-06
Add domain

bergsjoh:


bootstrap.py
```diff
@@ -82,6 +82,7 @@ def get_architecture():
 parser.add_option("-u", "--update", dest="update", action="store_true", help="Fully Updates the System")
 parser.add_option("-v", "--verbose", dest="verbose", action="store_true", help="Verbose output")
 parser.add_option("-f", "--force", dest="force", action="store_true", help="Force registration (will erase old katello and puppet certs)")
+parser.add_option("-d", "--add-domain", dest="add_domain", action="store_true", help="Automatically add the clients domain to Satellite6")
```
evgeni: s/Satellite6/Foreman/

also, I think I would only add the long `--add-domain` flag, `-d` sounds too much like debug or something and I think we do not need a short flag for everything-

bootstrap.py
```diff
@@ -369,13 +370,29 @@ def return_puppetenv_for_hg(hg_id):
...
+def create_domain(domain, orgid, locid):
+    myurl = "https://" + options.foreman_fqdn + ":" + API_PORT + "/api/v2/domains"
+    domid = return_matching_foreman_key('domains', 'name="%s"' % domain, 'id', True)
+    if not domid:
+        jsondata = json.loads('{"domain": {"name": "%s", "organization_ids": [%s], "location_ids": [%s]}}' % (domain, orgid, locid))
+        print_running("Calling Foreman API to create domain %s associated with the org & location" % domain)
+        post_json(myurl, jsondata)
+        return domain
+    else:
+        return domain
```
evgeni: both returns are not used, so could be dropped?

bootstrap.py
```diff
@@ -369,13 +370,29 @@ def return_puppetenv_for_hg(hg_id):
...
 
 def create_host():
     myhgid = return_matching_foreman_key('hostgroups', 'title="%s"' % options.hostgroup, 'id', False)
     mylocid = return_matching_foreman_key('locations', 'title="%s"' % options.location, 'id', False)
     myorgid = return_matching_foreman_key('organizations', 'name="%s"' % options.org, 'id', False)
     if DOMAIN:
-        mydomainid = return_matching_foreman_key('domains', 'name="%s"' % DOMAIN, 'id', False)
+        if options.add_domain:
+            create_domain(DOMAIN, myorgid, mylocid)
+        try:
```
evgeni: I think I'd prefer:

```
mydomainid = return_matching_foreman_key('domains', 'name="%s"' % DOMAIN, 'id', True)
if not mydomainid:
    print_generic("Domain %s doesn't exist in Foreman, consider using the --add-domain option." % DOMAIN)
    sys.exit(2)
```

bootstrap.py
bergsjoh: Originally had an idea to check ret value, but didn't turn out that way, so yeah it's not used anymore, removing in next push.

evgeni:
thanks, code LGTM :+1:, didn't run it though.

evgeni:
btw, Foreman will automatically create the domain from Puppet facts too. It's just too late for us :)

PR #87, opened 2016-06-07
don't try to delete the host from Foreman while --skip-foreman --remove

evgeni:
otherwise we do not honor what the user actually requested


sideangleside:
LGTM. Merging. Thanks @evgeni

PR #91, opened 2016-06-15
Added documentation, comments and reshuffled code with __main__ check

ericzolf:
Add documentation and reshuffled code with **main** check for pydoc usage.
Add comments with double hash (##) to better understand flow of script (that was actually a learning activity for me).
Use `pydoc ./bootstrap.py` to get the documentation.
Use `awk -F\#\# 'NF>1 {print $2}' ./bootstrap.py` to see the flow.
The commit is relatively important but if you look at the diff, you'll see that the code actually doesn't change, with the exception of the **main** check, and that only lines have been moved around and comments / documentation added all over the place.

There are a few TODO marks where I was not sure what the purpose / reason / context of the commented code was supposed to be.

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -1,5 +1,10 @@
 #!/usr/bin/python
-#
+"""
+script to register a new host to Foreman/Satellite or move it from Sat 5 to 6
+
+Use `pydoc ./bootstrap.py` to get the documentation.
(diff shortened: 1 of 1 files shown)


ericzolf:
Sorry, made a mistake regarding blanks vs. tabs...

Issue #101, opened 2016-07-21
document the need for a "recent" subsciption-manager-migration


evgeni:
in the case of migration, we call `rhn-migrate-classic-to-rhsm` with `--activation-key`, but that is only available since 1.14 (see https://github.com/candlepin/subscription-manager/commit/00461f3751f9db182227c9973c41b305e378638a)

this failed for me on a rhel 6.6 machine which only had 1.12 :(

PR #103, opened 2016-07-27
Keeping subscription-manager* packages to latest version

upadhyeammit:
When subscription-manager-migration\* packages are not latest migration can fail with errors as below, this patch checks if packages installed, if yes then updates those,

[ERROR], [2016-07-11 12:22:06], EXITING: [/usr/sbin/rhn-migrate-classic-to-rhsm --org Some_Org --activation-key some_ak  --destination-url=https://server.example.com:443/rhsm --legacy-user 'user' --legacy-password 'user-pass' --force] failed to execute properly.
Usage: rhn-migrate-classic-to-rhsm [OPTIONS]

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -177,9 +177,16 @@ def exec_failexit(command):
 
 
 def install_prereqs():
-    print_generic("Installing subscription manager prerequisites")
+    print_generic("Installing or updating subscription manager prerequisites")
     exec_failexit("/usr/bin/yum -y remove subscription-manager-gnome")
-    exec_failexit("/usr/bin/yum -y install subscription-manager subscription-manager-migration-*")
+    if os.path.exists('/usr/sbin/subscription-manager'):
+        exec_failexit("/usr/bin/yum -y update subscription-manager")
+    else:
+        exec_failexit("/usr/bin/yum -y install subscription-manager")
+    if os.path.exists('/usr/sbin/rhn-migrate-classic-to-rhsm'):
+        exec_failexit("/usr/bin/yum -y update subscription-manager-migration-*")
+    else:
+        exec_failexit("/usr/bin/yum -y install subscription-manager-migration-*")
     exec_failexit("/usr/bin/yum -y update yum openssl")
 
 


evgeni:
hey, thanks for the patch. but to my knowledge yum will already update a package if it is asked to install an installed one.

tested on EL6:

```
[root@c62c6bd64268 /]# yum install yum
Loaded plugins: fastestmirror, ovl
Setting up Install Process
Determining fastest mirrors
 * base: ftp.wrz.de
 * extras: ftp.halifax.rwth-aachen.de
 * updates: ftp.wrz.de
Resolving Dependencies
--> Running transaction check
---> Package yum.noarch 0:3.2.29-73.el6.centos will be updated
---> Package yum.noarch 0:3.2.29-75.el6.centos will be an update
--> Finished Dependency Resolution

Dependencies Resolved

===============================================================================================================================================================================
 Package                             Arch                                   Version                                              Repository                               Size
===============================================================================================================================================================================
Updating:
 yum                                 noarch                                 3.2.29-75.el6.centos                                 updates                                 1.0 M

Transaction Summary
===============================================================================================================================================================================
Upgrade       1 Package(s)

Total download size: 1.0 M
Is this ok [y/N]: n
Exiting on user Command
```

(same on EL5 and EL7)

upadhyeammit:
Hello,

My mistake, I should have handled this specifically for activation key. I can see that if have subscription-manager-migration package less than subscription-manager-migration-1.14.10-1.el6 version then /usr/sbin/rhn-migrate-classic-to-rhsm fails as prior versions do not have activation key support.

So should we handle this in 'install_prereqs' method ? I think we should put message that activation key not supported and we should ask for username and password for subscription-manager registration, as I can see we are doing it in below file,

/usr/share/rhsm/subscription_manager/migrate/migrate.py

```
    if self.options.activation_keys:
        self.destination_creds = UserCredentials(None, None)
    elif not self.is_hosted or self.options.destination_url or self.options.registration_state == "keep":
        self.destination_creds = self.authenticate(self.options.destination_user, self.options.destination_password,
            _("Destination username: "), _("Destination password: "))
    else:
        self.destination_creds = self.legacy_creds   # make them the same
```

If you agree I would submit patch again.
_
Amit Upadhye.

evgeni:
Hi Amit,

yes, this problem is "documented" in issue #101 
I tried to add proper error messages in pr #102 

given that we call `yum -y install subscription-manager subscription-manager-migration-*`, I would expect that yum actually already tries to upgrade those packages. did it not work for you?

I'd prefer not to use credentials here, as this would probably attach different subs than expected by the user.

@mccun934 @sideangleside opinions?

sideangleside:
Two options: 
1. Gracefully fail. 
2. We can treat this like we treat RHEL5, which also doesn't have a version of `rhn-classic-migrate-to-rhsm` that supports activation keys. That is, copy the .PEM file ourselves, then just run `subscription-manager` with the `--activation-key` switch. 

I do **NOT**  want to handle running `rhn-classic-migrate-to-rhsm` and handling credentials with it because
- The user doesn't get a whole lot of choice in what subscriptions get attached. (basically all we can do is run --auto-attach). 
- We'd have a second migration workflow.

upadhyeammit:
Hello,

https://github.com/Katello/katello-client-bootstrap/pull/102/commits/3421b2364d680557fccd53be8eba4a354229a6c4

Fixes handling of activation key, so all good.

evgeni:
ok, can this be closed then? or is there anything left to clean up?

upadhyeammit:
yes, we can close this one.

PR #107, opened 2016-09-18
Install Foreman's SSH key for remote execution.

fcami:
Install Foreman's SSH key for remote execution.
Default: false. Tested on RHEL7.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,4 +130,6 @@ Options:
   --unmanaged           Add the server as unmanaged. Useful to skip
                         provisioning dependencies.
   --rex                 Install Foreman's SSH key for remote execution.
+  --rex-user=REMOTE_EXEC_USER
+                        Local user used by Foreman's remote execution feature.
 ~~~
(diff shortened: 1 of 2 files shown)


sideangleside (commented):
Looks good so far. A few recommendations. 
- You might want to add a switch to allow a user-configurable user for REX. As per the [docs](https://www.theforeman.org/plugins/foreman_remote_execution/0.1/index.html#4.2Globalsettings), a user can configure which account is used via Foreman for SSH. We'd want to offer them similar with bootstrap.py
- As it is not assumed that the rex user is root, we'd have to place the public key in ~$REX_USER/.ssh/authorized_keys, so your code would need to be modified to handle that use-case. 
- Wrap your urllib2.urlopen calls in a try/except clause as you'd want to handle errors, be they network errors, or https://url:9090/ssh/pubkey not existing (as they would on older versions or if the user explicitly disables REX on the server)

fcami:
Understood, I will post an updated patch this week.

CalvinHartwell:
+1 for Frenchie! Code looks good, I like the name param 'rex'

Adding to Rich's feedback:
- Error handling would be a plus, even a simple try/except which handles and prints the http error code and the exception in a raw way would be good.

fcami:
- added --rex-user
- added try/except to catch urllib2 errors

PR #111, opened 2016-09-22
generate katello.facts to always pass fqdn in rhsm requests

evgeni:


sideangleside:
LGTM. Merging. Thanks @evgeni

PR #112, opened 2016-10-05
Enable subscription-manager yum plugins

sideangleside:
This PR ensures that the following yum plugins are enabled in subscription-manager prior running the register or migration routines. 
- the **product-id** plugin - ensures that new product certificates are placed in /etc/pki/product. Without this, systems won't get the product certs when products are installed from the various Red Hat repositories. 
- the **subscription-manager** plugin - ensures that subscription-manager works properly.

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -498,7 +498,6 @@ def disassociate_host(host_id):
 def configure_subscription_manager():
     configparser = SafeConfigParser()
     configparser.read('/etc/yum/pluginconf.d/product-id.conf')
-    configparser.get('main', 'enabled')
     if configparser.get('main','enabled') == '0':
       print_generic("Product-id yum plugin was disabled. Enabling...")
       configparser.set('main', 'enabled', '1')
@@ -507,7 +506,6 @@ def configure_subscription_manager():
 
     configparser = SafeConfigParser()
     configparser.read('/etc/yum/pluginconf.d/subscription-manager.conf')
-    configparser.get('main', 'enabled')
     if configparser.get('main','enabled') == '0':
       print_generic("subscription-manager yum plugin was disabled. Enabling...")
       configparser.set('main', 'enabled', '1')
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -496,21 +496,19 @@ def disassociate_host(host_id):
     put_json(myurl)
 
 def configure_subscription_manager():
-    configparser = SafeConfigParser()
-    configparser.read('/etc/yum/pluginconf.d/product-id.conf')
-    if configparser.get('main','enabled') == '0':
+    productidconfig = SafeConfigParser()
+    productidconfig.read('/etc/yum/pluginconf.d/product-id.conf')
+    if productidconfig.get('main','enabled') == '0':
       print_generic("Product-id yum plugin was disabled. Enabling...")
-      configparser.set('main', 'enabled', '1')
-      with open('/etc/yum/pluginconf.d/product-id.conf','w') as configfile:
-        configparser.write(configfile)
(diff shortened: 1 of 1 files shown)


bootstrap.py
```diff
@@ -495,6 +495,22 @@ def disassociate_host(host_id):
     print_running("Disassociating host id %s for host %s" % (host_id, FQDN))
     put_json(myurl)
 
+def configure_subscription_manager():
+    configparser = SafeConfigParser()
+    configparser.read('/etc/yum/pluginconf.d/product-id.conf')
+    if configparser.get('main','enabled') == '0':
+      print_generic("Product-id yum plugin was disabled. Enabling...")
+      configparser.set('main', 'enabled', '1')
+      with open('/etc/yum/pluginconf.d/product-id.conf','w') as configfile:
```
evgeni: `with` is Python 2.5+ only. :cry:
can EL5 please die?

bootstrap.py
```diff
@@ -495,6 +495,22 @@ def disassociate_host(host_id):
     print_running("Disassociating host id %s for host %s" % (host_id, FQDN))
     put_json(myurl)
 
+def configure_subscription_manager():
+    configparser = SafeConfigParser()
```
evgeni: `configparser` sounds so generic, why not `productidconfig`?

bootstrap.py
```diff
@@ -495,6 +495,22 @@ def disassociate_host(host_id):
...
+    configparser.read('/etc/yum/pluginconf.d/product-id.conf')
+    if configparser.get('main','enabled') == '0':
+      print_generic("Product-id yum plugin was disabled. Enabling...")
+      configparser.set('main', 'enabled', '1')
+      with open('/etc/yum/pluginconf.d/product-id.conf','w') as configfile:
+        configparser.write(configfile)
+
+    configparser = SafeConfigParser()
```
evgeni: `configparser` sounds so generic, why not `submanconfig`?

bootstrap.py
```diff
@@ -495,6 +495,22 @@ def disassociate_host(host_id):
...
+    configparser.read('/etc/yum/pluginconf.d/subscription-manager.conf')
+    if configparser.get('main','enabled') == '0':
+      print_generic("subscription-manager yum plugin was disabled. Enabling...")
+      configparser.set('main', 'enabled', '1')
+      with open('/etc/yum/pluginconf.d/product-id.conf','w') as configfile:
```
evgeni: `configfile` sounds so generic, why not `submanfile`?
also: wrong file!

bootstrap.py
evgeni: `configfile` sounds so generic, why not `productidfile`?

evgeni (changes requested):

evgeni (approved):

sideangleside:
Added some updates in commit https://github.com/Katello/katello-client-bootstrap/pull/112/commits/334cfba45035863863602bdc15f3b7c7aa495338 based on feedback

evgeni:
@sideangleside yupp, LGTM now.

bootstrap.py
bergsjoh: 6 months my friend, 6 months 👍

evgeni: @bergsjoh partially true, production 3 is then over, but people will still have it...

Rocco83: Will have for a long time, i can ensure... 
I know that there are still RHEL4 and RHEL3 in prod.

evgeni:
merged in 7f84e1afe5cca26927aa2aafdcab3d5235c973d9

PR #113, opened 2016-10-06
update python to avoid gopher issues (fixes RHBZ #1360360)

sideangleside:
From [bz1360360](https://bugzilla.redhat.com/show_bug.cgi?id=1360360):

```
Python on RHEL 6.4 and older is affected by https://bugzilla.redhat.com/show_bug.cgi?id=845802
This leads to a lot of message spam on the consoles when goferd is running.
bootstrap.py should try to upgrade python (as it does with yum, openssl etc) to avoid this issue
```


evgeni (approved):

evgeni:
Why didn't I open that PR ages ago? :)

PR #114, opened 2016-10-10
Add --enablerepo option to support older versions of subscription-manager

sideangleside:
As part of commit https://github.com/Katello/katello-client-bootstrap/commit/1912d60eac04cba8288b06b2273d02784d28b88a,  the `enable_sat_tools` function, which explicitly enabled the Satellite Tools repositories was removed. 

This effectively meant that users are required to use a version of `subscription-manager` > 1.10 which enables / disables repository based upon the Product Content attributes of an activation key. 

Some users of particularly old versions of RHEL5 & RHEL6 do not (and cannot) have a version of subscription-manager new enough to support Product Content attributes on activation keys. This made the process of running bootstrap.py on these older systems difficult (as the required repos to install puppet or katello-agent were not enabled)

The feature adds the `--enablerepos` switch which allows a user to pass a list of repositories, comma separated (and wildcards OK) which will be enabled via `subscription-manager`

Example Usage:

```
#./bootstrap.py -l admin \
  -s server.example.com \
  -o Default \
  -g RHEL7 \
  --enablerepos *satellite*,rhel-5-server-rpms
```

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -83,6 +83,7 @@ parser.add_option("--no-remove-obsolete-packages", dest="removepkgs", action="st
 parser.add_option("--unmanaged", dest="unmanaged", action="store_true", help="Add the server as unmanaged. Useful to skip provisioning dependencies.")
 parser.add_option("--rex", dest="remote_exec", action="store_true", help="Install Foreman's SSH key for remote execution.", default=False)
 parser.add_option("--rex-user", dest="remote_exec_user", default="root", help="Local user used by Foreman's remote execution feature.")
+parser.add_option("--enablerepos", dest="enablerepos", help="Repositories to be enabled via subscription-manager - comma separated", metavar="enablerepos")
 (options, args) = parser.parse_args()
 
 if not (options.foreman_fqdn and options.login and (options.remove or (options.org and options.activationkey and (options.no_foreman or options.hostgroup)))):
@@ -503,6 +504,12 @@ def check_rhn_registration():
     else:
         return False
 
+def enable_repos():
+    for repo in options.enablerepos.split(','):
+      print_running("Enabling repo - %s" % repo)
+      exec_failok("/usr/sbin/subscription-manager repos --enable %s" % repo)
+
+
 
 def get_api_port():
     configparser = SafeConfigParser()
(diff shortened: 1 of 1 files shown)


evgeni:
There is a patch downstream that adds `enable_sat_tools` back to the Satellite version of bootstrap. So it's actually not that a big issue.

But I like the idea and would love to adopt the downstream patch to set a default `enablerepos` to `rhel-*-satellite-tools-*-rpms` after merging this.

:+1:

bootstrap.py
```diff
@@ -503,6 +504,12 @@ def check_rhn_registration():
     else:
         return False
 
+def enable_repos():
+    for repo in options.enablerepos.split(','):
+      print_running("Enabling repo - %s" % repo)
+      exec_failok("/usr/sbin/subscription-manager repos --enable %s" % repo)
```
evgeni: IMHO we do not need to call tools with a full path.

evgeni: `sub-man repos --enable <foo>` is quite expensive.
Instead of calling it once for every repo requested, I would suggest to run it once with `--enable foo1 --enable foo2 …`

``` python
>>> repos = "foo,bar,baz"
>>> " ".join(['--enable=%s' % i for i in repos.split(',')])
'--enable=foo --enable=bar --enable=baz'
```

evgeni (changes requested):

sideangleside:
Should we even set a default `enablerepos` variable? There are users who create content views that have multiple OS versions and/or release versions of the tools repo (this is even recommended in some of the docs/guides). If we enable `rhel-*-satellite-tools-*-rpms` we'd enable conflicting versions of the tools repo and the user might not be able to successfully bootstrap the system. See [RHBZ #1378339](https://bugzilla.redhat.com/show_bug.cgi?id=1378339) for an example of this. 

I am of the mindset that since the latest release of RHEL 5, 6, & 7  (5.11, 6.8 & 7.2) all ship `subscription-manager` >= 1.10, enabling repos via the Product Content attribute on the activation key will be the default for almost all the users.  `--enablerepos` would really only be for 'advanced usage' where the user has an older release that they wish to migrate with minimal change on the system. (and they know what they are doing)

evgeni:
Mhh, funny BZ. But that's something that we should consider downstream. Your PR here is useful regardless of what we decide there.

evgeni (approved):

evgeni:
LGTM :+1:

PR #116, opened 2016-10-13
Fix legacy opts

sideangleside:
if the user passes `--legacy-purge`, we need to prompt them for `--legacy-password`if they haven't provided it via CLI parameter. 

Otherwise, the subsequent call to `rhn-migrate-classic-to-rhsm` in the `migrate_systems` function will fail.

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -100,7 +100,7 @@ if not DOMAIN and not (options.force or options.no_puppet):
 if not options.password and not options.no_foreman:
     options.password = getpass.getpass("%s's password:" % options.login)
 
-if options.legacy_login and not options.legacy_password:
+if options.legacy_purge and not options.legacy_password:
     options.legacy_password = getpass.getpass("Legacy User %s's password:" % options.legacy_login)
 
 if options.no_foreman:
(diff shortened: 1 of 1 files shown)


evgeni:
mhh, I _thought_ we had that covered, but apparently not.

not strictly like that you echo the password in clear-text again, but looks fine enough otherwise.

sideangleside:
For what it's worth, we do the same for the foreman user's password, and _only_ in verbose mode.

evgeni:
true that. yeah.

evgeni:
shall we also promt for legacy_login? currently it can be empty…

sideangleside:
Our default is to pass the `--keep` switch to `rhn-migrate-classic-to-rhsm`, which doesn't require a login to the legacy system (RHN Classic, Sat5, Spacewalk). This allows the user
- the ability to migrate systems that he may have inherited, but does not (for whatever reason) have a login to the legacy platform. 
- the ability to keep the legacy profile for audit/historical purposes (and as a fall back if they choose to move the system back to the old platform). 

Only`--purge` requires a login to the legacy system & we default legacy_login to **admin**, so I don't think it necessary to prompt for legacy_login in all cases.

evgeni:
Right. Remind me next time to _first_ have coffee, then review code :-)
LGTM then.

PR #121, opened 2016-10-31
Fix --skip-foreman to not require username/password

sideangleside:
Fixes the --skip-foreman switch to not require username/password. Many users leverage bootstrap to register for content only, and we don't need to prompt for username/password in those cases. 

One side-effect of this PR is that when leveraging the `--skip-foreman` switch, we cannot map organization *name* -> organization *label* as we are no longer using the API, and the user needs to provide the organization via label. 

Updated `--help` output and added **EXAMPLES.md** to illustrate various example uses.

### DIFF
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -1,0 +1,44 @@
+## Example usages of bootstrap.py
+
+
+### Registering a system to Foreman + Katello
+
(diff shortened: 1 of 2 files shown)


evgeni:
LGTM, just tested it in my 6.2.3 lab, works fine.

PR #122, opened 2016-11-02
Update yum functions to properly escape wildcards. 

sideangleside:
In the use-case when the user has a file named `katello-ca-consumer*` in the directory which bootstrap is invoked from, the script will fail to run when using the `--force` option as the function calls `yum remove katello-ca-consumer*` which matches the file on disk, and not the one(s) in the RPM database. 

This PR updates bootstrap to ~~query the RPM database for packages named katello-ca-consumer* and pass those as explicit arguments to the `yum remove` command~~ properly quote any call to yum which has a wildcard.

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -201,7 +201,7 @@ def check_migration_version():
 def install_prereqs():
     print_generic("Installing subscription manager prerequisites")
     yum("remove", "subscription-manager-gnome")
-    yum("install", "subscription-manager subscription-manager-migration-*")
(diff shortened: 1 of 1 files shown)


evgeni:
I'd prefer if we would use `subprocess` instead, without a shell…
Or maybe quote the stuff we pass?

Either way, I'd love not to touch rpmdb if not really needed.

sideangleside:
Good point. Updated the title (and code) to simply escape the wildcards so the shell doesn't interpret them.

PR #127, opened 2016-11-10
Fix unquoted activation key in migrate_systems

sideangleside:
The `migrate_systems()` function wasn't properly quoting the activation key parameter when call `rhn-classic-migrate-to-rhsm,` causing migrations to fail if the AK has spaces.


evgeni:
activation keys can have spaces?! o_O

but LGTM!

PR #120, opened 2016-10-28
Add check and graceful failure if katello-ca-consumer package is already installed. 

sideangleside:
Adds a more graceful message informing the user to run the script with the `--force` option in the event a `katello-ca-consumer-*` package is installed. Fixes #119

PR #128, opened 2016-11-14
remove Katello CA certs in clean_katello_agent

sideangleside:
In https://github.com/Katello/katello-client-bootstrap/pull/120, we added support for gracefully failing in the event that a system was already registered. (Done via checking for the presence of `/etc/rhsm/ca/katello-server-ca.pem`) As newer versions of the katello-ca-consumer RPM no longer deploy `/etc/rhsm/ca/katello-server-ca.pem` directly, but call `katello-rhsm-consumer`, which deploys the file, the **clean_katello_agent** function needs to delete *both* the `/etc/rhsm/ca/katello-server-ca.pem` file AND the RPM.

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -94,6 +94,27 @@ def exec_failexit(command):
     print_success(command)
     print ""
 
+def delete_file(filename):
+    """Helper function to delete files."""
+    try:
+        os.remove(filename)
+        print_success("Removing %s" % filename)
+    except OSError, e:
+        print_generic("Error when removing %s - %s" % (filename, e.strerror))
+        print_error("Removing %s" % filename)
+        sys.exit(1)
+
+def delete_directory(directoryname):
+    """Helper function to delete directories."""
+    try:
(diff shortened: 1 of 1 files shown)


evgeni (changes requested):
https://docs.python.org/2/library/os.html#os.unlink

sideangleside:
Should we also remove the `rm -rf` in the `clean_puppet()` function? I considered going with a pure python approach, but used `rm` here to be consistent. I am fine with either approach.

evgeni:
Yeah, we probably should go all Python here. But in the puppet case we will need https://docs.python.org/2/library/shutil.html#shutil.rmtree

sideangleside:
Added two new helper functions,`delete_file()` & `delete_directory()` to support file and directory deletion and retrofitted our current file and directory deletions to use them. 

I suspect that as bootstrap evolves, we will need to delete additional files and directories, so it would be helpful to have proper functions, with our custom notification messages and the like.

PR #130, opened 2016-11-16
Options for PE and RHEL-H added

fftux:
Options were added to configure systems to use puppet enterprise (v3) instead of integrated puppet in katello.  Additional mods may need to be made to be v4 compliant.  

Option added to configure virt-who on RHEL based systems that are configured as RHV/oVirt hypervisors.


evgeni:
Ohai,

I will add some inline comments, but here are a few general ones :-)
- can you sqash the changes in a few logical commits please?
- I don't think doing virt-who is bootstrap.py's job, we have Puppet/Ansible for that (but @sideangleside and @mccun934 might think otherwise)

bootstrap.py
```diff
@@ -232,6 +236,11 @@ def register_systems(org_name, activationkey, release):
 def unregister_system():
     print_generic("Unregistering")
     exec_failexit("/usr/sbin/subscription-manager unregister")
+    
+    
+def enable_satellite_tools():
```
evgeni, line 241: that's a downstream Satellite thing, now replaced with `--enablerepos` upstream.

bootstrap.py
```diff
@@ -288,11 +301,24 @@ def install_puppet_agent():
     exec_failexit("/sbin/service puppet restart")
 
 
+def install_pe_agent(org_name):
+    org_label = return_matching_katello_key('organizations', 'name="%s"' % org_name, 'label', False)
+    if os.path.isfile('/etc/yum.repos.d/pe_repo.repo'):
+        exec_failexit("mv /etc/yum.repos.d/pe_repo.repo{,.bak}")
+    exec_failexit("subscription-manager attach --pool=$(subscription-manager list --available --matches='Puppet Enterprise' --pool-only)" )
```
evgeni, line 308: the pool belongs into the activation-key, not into our calls

bootstrap.py
```diff
@@ -288,11 +301,24 @@ def install_puppet_agent():
...
+    exec_failexit("subscription-manager repos --disable=*-pe-*")
+    exec_failexit("subscription-manager repos --enable=%s_Puppet_Enterprise_$(uname -r | cut -d. -f6)-pe-$(uname -r | cut -d. -f7)" % ( org_label ))
```
evgeni, line 310: those two should be done using `--enablerepos` as we don't know how the repos will be called at the users setup

bootstrap.py
```diff
@@ -288,11 +301,24 @@ def install_puppet_agent():
...
+    print_generic("Installing the PE Agent")
+    yum("install", "pe-agent")
+    exec_failexit("/sbin/chkconfig pe-puppet on")
+    exec_failexit("/usr/local/bin/puppet agent -t --server %s || true" % (options.pe_server_fqdn))
```
evgeni, line 314: why `|| true`? this hides errors in puppet comms

bootstrap.py
```diff
@@ -288,11 +301,24 @@ def install_puppet_agent():
...
+    exec_failexit("/sbin/service pe-puppet restart")
+
+
 def remove_obsolete_packages():
     print_generic("Removing old RHN packages")
     yum("remove", "rhn-setup rhn-client-tools yum-rhn-plugin rhnsd rhn-check rhnlib spacewalk-abrt spacewalk-oscap osad rh-*-rhui-client")
 
-
```
evgeni, line 295: pep8 wants this :)

bootstrap.py
```diff
@@ -323,6 +349,20 @@ def install_foreman_ssh_key():
         os.chown(foreman_ssh_authfile, userpw.pw_uid, userpw.pw_gid)
         print_generic("Foreman's SSH key was added to %s" % foreman_ssh_authfile)
 
+def install_virt_who():
```
evgeni, line 352: as mentioned, I don't think this belongs here

bootstrap.py
```diff
@@ -600,18 +640,29 @@ class MEOptions:
     register_systems(options.org, options.activationkey, options.release)
 
 if not options.remove:
+    enable_satellite_tools()
```
evgeni, line 643: Satellite stuff…

bootstrap.py
```diff
@@ -288,11 +301,24 @@ def install_puppet_agent():
     exec_failexit("/sbin/service puppet restart")
 
 
+def install_pe_agent(org_name):
+    org_label = return_matching_katello_key('organizations', 'name="%s"' % org_name, 'label', False)
+    if os.path.isfile('/etc/yum.repos.d/pe_repo.repo'):
```
evgeni, line 306: I _think_ I'd leave that to the user. The file might be called differently too.

bootstrap.py
```diff
@@ -288,11 +301,24 @@ def install_puppet_agent():
     exec_failexit("/sbin/service puppet restart")
 
 
+def install_pe_agent(org_name):
```
evgeni, line 304: can't we merge that and install_puppet? just having a flag whether it's puppet or PE that is going to be installed?

evgeni (changes requested):

evgeni:
Ah, and one more thing, we're missing a cleanup for PE installs.

bootstrap.py
sideangleside: +1 Attaching subs is the activation keys' job. Also, this  assumes the custom Product is named **Puppet Enterprise** . 

Also, file operations (creates, deletes, moves & removes) should use native Python modules to remain consistent.

bootstrap.py
sideangleside: Or done via the activation key. It is the user's choice, but -1 to hardcoded repo names.

sideangleside:
I don't think a function to install (and configure) virt-who is within the scope of what bootstrap is designed to do. I would be open to, as a courtesy to users who are using Katello solely as a content source and have no config mgmt, a generic function to install additional packages (provided in a comma separated fashion). After that, we are well into Puppet/Ansible territory.

bootstrap.py
sideangleside: Yes. Though the PE Agent requires a bit more configuration especially if a user is using Trusted Facts.

PR #123, opened 2016-11-07
disable plugin before removing

luigizhou:
Disable rhnplugin before removing, avoiding error in case the plugin is installed and enabled.

PR #136, opened 2016-11-30
move systemid when rhn-migrate-classic-to-rhsm was called with --keep

evgeni:
When rhn-migrate-classic-to-rhsm is called with --keep, it will leave the systemid
file intact, which might confuse the (not yet removed) yum-rhn-plugin.
Move the file to a backup name, so the user can still restore it if needed.

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -195,6 +195,12 @@ def migrate_systems(org_name, activationkey):
     exec_failexit("/usr/sbin/rhn-migrate-classic-to-rhsm --org %s --activation-key '%s' %s" % (org_label, activationkey, options.rhsmargs))
     exec_failexit("subscription-manager config --rhsm.baseurl=https://%s/pulp/repos" % options.foreman_fqdn)
 
+    # When rhn-migrate-classic-to-rhsm is called with --keep, it will leave the systemid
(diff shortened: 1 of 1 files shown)


sideangleside:
I assume this PR would effectively obviate the need to merge #123? or do we want both?

evgeni:
it's complicated...

#123 is a one-off disable of the plugin for the step when we remove the plugin (as its a bit racy). I think we should rather do something like https://github.com/candlepin/subscription-manager/blob/master/src/subscription_manager/migrate/migrate.py#L599 after moving away systemid. just for the case the user called --no-remove-packages, to save a few CPU cycles.

but the problem reported in #123 should be gone after this PR is merged

sideangleside:
I'd recommend disabling the plugin as above so as to not confuse the end-user. If you just remove the systemid file, you get 

~~~
yum repolist
Loaded plugins: product-id, rhnplugin, security, subscription-manager
This system may not be registered to RHN Classic or RHN Satellite. SystemId could not be acquired.
You can use rhn_register to register.
RHN Satellite or RHN Classic support will be disabled.
~~~

It doesn't matter because the plugin will be removed very shortly thereafter, but I can imagine it being confusing for the uninitiated.

evgeni:
closed in favour of #138

PR #137, opened 2016-11-30
call rhn-migrate-classic-to-rhsm with --remove-rhn-packages

evgeni:
but only if we have removepkgs set (which is the default, though)

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -89,6 +89,8 @@ ERROR_COLORS = {
 }
 
 SUBSCRIPTION_MANAGER_SERVER_TIMEOUT_VERSION = '1.18.2'
+SUBSCRIPTION_MANAGER_MIGRATION_MINIMAL_VERSION = '1.14.2'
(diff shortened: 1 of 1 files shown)


sideangleside (commented):
Not every version of rhn-migrate-classic-to-rhsm supports the --remove-rhn-packages option. see (https://bugzilla.redhat.com/show_bug.cgi?id=1185914) so I'd be a little uncomfortable calling it blindly. 

We could either check for an appropriate version and call it only in that use case or stick with our function that removes packages. The upside of our function is that it works in all uses (migration and registration)

evgeni:
bah, you are right. we'd have to wrap that into if version... bah bah bah

my idea was to keep both, as this allows us to benefit from the knowledge inside of migrate-to-rhsm

sideangleside:
It doesn't hurt to run both. I'm OK with this.

evgeni:
Just so that I don't loose the link: https://bugzilla.redhat.com/show_bug.cgi?id=1185914

beav:
@evgeni this PR needs a rebase

ehelms (approved):

PR #138, opened 2016-12-01
Properly disable RHN Plugin

sideangleside:
Added `disable_rhn_plugin()` function which will disable the RHN plugin by

- moving the systemid file to systemid-bak
- disabling the RHN plugin in `/etc/yum/pluginconf.d/rhnplugin.conf`.

Updated the `migrate_systems` and `prepare_rhel5_migration ` functions to use it. 

This will ensure the RHN plugin is properly disabled at the completion of a run of bootstrap. 

This PR obsoletes #136 & #123

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -174,6 +174,17 @@ def get_bootstrap_rpm():
     exec_failexit("rpm -Uvh http://%s/pub/katello-ca-consumer-latest.noarch.rpm" % options.foreman_fqdn)
 
 
+def disable_rhn_plugin():
+    rhnpluginconfig = SafeConfigParser()
+    rhnpluginconfig.read('/etc/yum/pluginconf.d/rhnplugin.conf')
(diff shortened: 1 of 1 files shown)


bootstrap.py
```diff
@@ -174,6 +174,17 @@ def get_bootstrap_rpm():
     exec_failexit("rpm -Uvh http://%s/pub/katello-ca-consumer-latest.noarch.rpm" % options.foreman_fqdn)
 
 
+def disable_rhn_plugin():
+    rhnpluginconfig = SafeConfigParser()
+    rhnpluginconfig.read('/etc/yum/pluginconf.d/rhnplugin.conf')
```
evgeni: what happens if the file is not present? `SafeConfigParser.read()` will survive, but the `get()`?
I think we should wrap this into `if os.path.exists('/etc/yum/pluginconf.d/rhnplugin.conf')`, no?

sideangleside: Ya.

PR #141, opened 2016-12-05
document required permissions

evgeni:


evgeni:
@sideangleside @MartijntenHeuvel i can haz review plz?

sideangleside (changes requested):
we'd also need to document the permissions needed when using the --legacy-purge option, which are 

User needs to be an admin of the system in Sat5/RHN by having any of the following roles:

- An organization administrator
- A system group administrator for a system group that the system is a member of. 
- granted permissions to the system explicitly via Users-> account-> 'Systems Administered by this User'

MartijntenHeuvel:
If you give the edit_hosts _role_ this will provide more then required: you can give only create_hosts instead of "view_hosts, create_hosts, edit_hosts, destroy_hosts, build_hosts" which is again more then required. 

you can also create a new role, add only the create_hosts to that; shouldn't that be sufficient?

evgeni:
@sideangleside @MartijntenHeuvel added more words, please re-review.

sideangleside:
:+1:

sideangleside (approved):

MartijntenHeuvel:
Looks good to me! +1 @evgeni

evgeni:
thanks guys, merging :)

PR #144, opened 2017-01-03
Check for /etc/rhsm/facts before fetching facts

dLobatog:
If the host does not have this directory because it 
never had subscription manager, bootstrap.py will fail.

This commit simply takes care of checking whether the
directory exists before opening katello.facts

Submitted for https://bugzilla.redhat.com/show_bug.cgi?id=1408972

Meant for https://github.com/Katello/katello-client-bootstrap/issues/143

PR #145, opened 2017-01-03
Clean environment after prereqs installation

### LINKED ISSUE #143: Creation of /etc/rhsm/facts/katello.facts fails on systems that do not have subscription-manager installed
On older systems that do not have subscription-manager (or on systems where the admin never installed it), the creation of `/etc/rhsm/facts/katello.facts fails` with a traceback similar to the following:

~~~
  File "./bootstrap.py.orig", line 797, in <module>
    clean_environment()
  File "./bootstrap.py.orig", line 277, in clean_environment
    katellofacts = open('/etc/rhsm/facts/katello.facts', 'w')
IOError: [Errno 2] No such file or directory: '/etc/rhsm/facts/katello.facts'
~~~

This is because `/etc/rhsm/facts` is created (and owned) by the subscription-manager package and doesn't exist until after the package is installed. 

We can address this one of two ways. 

- run the pythonic equivalent of `mkdir -p /etc/rhsm/facts` in the `clean_environment` function. 
- move the code that creates `/etc/rhsm/facts/katello.facts` to run after subscription-manager is installed in the `install_prereqs` function. 

See also [RH BZ1408972 ](https://bugzilla.redhat.com/show_bug.cgi?id=1408972)

dLobatog:
Since cleaning the environment relies on subscription manager
directories being available (/etc/rhsm/facts), we should run it after
the installation happens.

An alternative for https://github.com/Katello/katello-client-bootstrap/pull/144/files to fix Katello#143 and https://bugzilla.redhat.com/show_bug.cgi?id=1408972

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -156,6 +156,7 @@ def install_prereqs():
     yum("remove", "subscription-manager-gnome")
     yum("install", "subscription-manager 'subscription-manager-migration-*'")
     yum("update", "yum openssl python")
+    generate_katello_facts()
 
 
 def get_bootstrap_rpm():
@@ -191,7 +192,7 @@ def migrate_systems(org_name, activationkey):
     Call `rhn-migrate-classic-to-rhsm` to migrate the machine from Satellite
     5 to 6 using the organization name/label and the given activation key, and
     configure subscription manager with the baseurl of Satellite6's pulp.
-    This allows the administrator to override the URL provided in the 
+    This allows the administrator to override the URL provided in the
     katello-ca-consumer-latest RPM, which is useful in scenarios where the
     Capsules/Servers are load-balanced or using subjectAltName certificates.
     If called with "--legacy-purge", uses "legacy-user" and "legacy-password"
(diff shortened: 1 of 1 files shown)


bootstrap.py
```diff
@@ -156,6 +156,9 @@ def install_prereqs():
     yum("remove", "subscription-manager-gnome")
     yum("install", "subscription-manager 'subscription-manager-migration-*'")
     yum("update", "yum openssl python")
+    # > Clean the environment from LD_... variables
+    clean_environment()
```
evgeni: this is too late. clean_environment is there to fix up unclean environments as created by BMC BladeLogic. those tend to break yum. I'd argue that we should instead split out the fact generation out of clean_environement, and call that part here instead.

dLobatog:
How about this? Cleaning vars before installing and creating facts after the installation (when we know for sure that the directory exists?

evgeni:
I'd leave the clean_env call where it was (just for the sake of that is what I had tested at $customer when it occured). Otherwise exactly what I meant. Thanks!

dLobatog:
@evgeni Fixed as per your comments!

evgeni:
LGTM! :+1: 

@sideangleside whatcha think?

evgeni (approved):

sideangleside:
LGTM 👍

dLobatog:
@evgeni @sideangleside I can't merge this myself 😄

evgeni:
Right. Sorry. Meant to do that yesterday, but then forgot.

PR #147, opened 2017-01-06
Make sure rhsmcertd is enabled. Some users may have disabled while us…

d-kaylor:
…ing RHN.

Installing subscription-manager enables rhsmcertd but I ran into a case where a user had disabled rhsmcertd while using RHN and wondered why checkins were not occurring in Satellite 6.

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -156,6 +156,7 @@ def install_prereqs():
     yum("remove", "subscription-manager-gnome")
     yum("install", "subscription-manager 'subscription-manager-migration-*'")
     yum("update", "yum openssl python")
+    exec_failexit("/sbin/chkconfig rhsmcertd on")
     generate_katello_facts()
 
 
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -156,7 +156,6 @@ def install_prereqs():
     yum("remove", "subscription-manager-gnome")
     yum("install", "subscription-manager 'subscription-manager-migration-*'")
     yum("update", "yum openssl python")
-    exec_failexit("/sbin/chkconfig rhsmcertd on")
     generate_katello_facts()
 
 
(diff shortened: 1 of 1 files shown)


evgeni:
makes sense, yeah. I would prefer this would happen *after* migrate_systems/register_systems, though. and it should be probably also actually (re)started then?

sideangleside:
rhsmcertd gets enabled and restarted on install of subscription-manager 
~~~
rpm -q --scripts subscription-manager
postinstall scriptlet (using /bin/sh):
    /bin/systemctl enable rhsmcertd.service >/dev/null 2>&1 || :
    /bin/systemctl daemon-reload >/dev/null 2>&1 || :
    /bin/systemctl try-restart rhsmcertd.service >/dev/null 2>&1 || :
~~~
and restarted on upgrades. 

~~~
postuninstall scriptlet (using /bin/sh):
    
systemctl daemon-reload >/dev/null 2>&1 || : 
if [ $1 -ge 1 ] ; then 
        # Package upgrade, not uninstall 
        systemctl try-restart rhsmcertd.service >/dev/null 2>&1 || : 
fi
~~~

@d-kaylor: other than the lack of updated check-ins, were there any other ill-effects of having rhsmcertd disabled & not running while running bootstrap? If no, then I am in agreement with @Zhenech that the daemon should be (re)started/enabled _after_ migrate_systems/register_systems.

d-kaylor:
There were no problems during the run of bootstrap so moving the service restart & re-enable should be fine.

evgeni:
looks good code-wise, ENOTIME to test it, though. sorry.

maybe @sideangleside can?

sideangleside:
I'll test and report back. Cursorily, it looks good to me.

sideangleside:
LGTM :+1: Merging. Thanks @d-kaylor

PR #152, opened 2017-01-31
fix evaluation of the retcode when using commands.getstatusoutput

evgeni:
commands.getstatusoutput() returns the exit code that should be
"interpreted according to the rules for the C function wait()."[1]

However, wait() does not return the exit code only, but also other
information how the process ended (like the signal). To get the exit
code, wait(3)[2] documents a macro:

 WEXITSTATUS(stat_val)
   If the value of WIFEXITED(stat_val) is non-zero, this macro
   evaluates to the low-order 8 bits of the status argument that
   the child process passed to _exit() or exit(), or the value the
   child process returned from main().

This means that we need not the whole exit status from
getstatusoutput(), but only the lower 8 bits. Get these by shifting
'result' by 8 bits to the right.

[1] https://docs.python.org/2/library/commands.html#commands.getstatusoutput
[2] https://linux.die.net/man/3/wait

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -73,7 +73,7 @@ def exec_failok(command):
     """Helper function to call a command with only warning if failing."""
     print_running(command)
     output = commands.getstatusoutput(command)
-    retcode = output[0]
+    retcode = output[0] >> 8
     if retcode != 0:
         print_warning(command)
     print output[1]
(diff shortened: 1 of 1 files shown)


sideangleside:
:+1: Merging

PR #158, opened 2017-02-16
make katello-ca-consumer RPM download configurable to use HTTP or HTTPS

### LINKED ISSUE #97: [RFE] Make katello-ca-consumer rpm download over https
Reduce the requirements of firewall openings even more.

Keeping port 80 opened to capsule/satellite as a requirement just to download the bootstrap from clients seems a bit unnecessary.

sideangleside:
Fixes #97. Bootstrap now defaults to using HTTPS to download the `katello-ca-consumer` package.

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -163,7 +163,7 @@ def install_prereqs():
 def get_bootstrap_rpm():
     """
     Retrieve Client CA Certificate RPMs from the Satellite 6 server.
-    Uses --insecure options to curl(1) to download via HTTPS
+    Uses --insecure options to curl(1) if instructed to download via HTTPS
     If called with --force, calls clean_katello_agent().
     """
     if options.force:
@@ -176,7 +176,7 @@ def get_bootstrap_rpm():
     if options.download_method == "https":
         print_generic("Writing custom cURL configuration to allow download via HTTPS without certificate verification")
         curl_config_dir = tempfile.mkdtemp()
-        curl_config = open(("%s/.curlrc" % curl_config_dir),'wb')
+        curl_config = open((os.path.join(curl_config_dir, '.curlrc'), 'wb')
         curl_config.write("insecure")
         curl_config.close()
         os.environ["CURL_HOME"] = curl_config_dir
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -173,17 +173,21 @@ def get_bootstrap_rpm():
         print_generic("To override this behavior, run the script with the --force option. Exiting.")
         sys.exit(1)
 
-    print_generic("Writing custom Curl configuration to allow download via HTTPS without certificate verification")
-    curl_config_dir = tempfile.mkdtemp()
-    curl_config = open(("%s/.curlrc" % curl_config_dir),'wb')
-    curl_config.write("insecure")
-    curl_config.close()
-    os.environ["CURL_HOME"] = curl_config_dir
-    print_generic("Retrieving Client CA Certificate RPMs")
-    exec_failexit("rpm -Uvh https://%s/pub/katello-ca-consumer-latest.noarch.rpm" % options.foreman_fqdn)
-    print_generic("Deleting Curl configuration")
-    delete_directory(curl_config_dir)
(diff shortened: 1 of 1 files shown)


evgeni:
ACK on the idea
NACK on unconditionally using HTTPS because Capsules currently don't have /pub on HTTPS (see https://bugzilla.redhat.com/show_bug.cgi?id=1370582 and http://projects.theforeman.org/issues/16904)

sideangleside:
Assuming that BZ is resolved, any other issues? We can hold off on merging until closure of that BZ.

evgeni:
* cURL is spelled cURL, not Curl ;)
* I wonder if we want a http switch for people using old Capsules with new bootstrap (but probably not)?

sideangleside:
~~~
cURL is spelled cURL, not Curl ;)
~~~

Yep. Will respect the brand and update:)

~~~
I wonder if we want a http switch for people using old Capsules 
with new bootstrap (but probably not)?
~~~

In Satellite, the `katello-client-bootstrap` package is delivered as part of the Satellite/Capsule build. Thus the version shipped would be compatible with your Satellite/Capsule. The people a `--http` would help are those who use the latest upstream against older downstream builds. Alternatively making the download method configurable would make this implementation no longer dependant on Red Hat BZ1370582.

evgeni:
Yeah, for pure Satellite users this should be fine. It's just me who usually just grabs the GitHub copy directly instead of taking the one from Satellite ;)

I like the idea of having `--download-method`, default it to http for now, flip to https when all the stack is happy.

sideangleside:
updated to include the `--download-method` option. Renaming this PR to reflect the change in direction.

evgeni:
Looks good, yeah. I wonder if Python 2.4 already supports "choices", so we can limit the methods to http and https (so people don't try ftp and gopher!)

sideangleside:
Well, I only use https if you specify it. otherwise, you get http. So a user who passes a non-sensical value gets http :) I am scared to look at Python 2.4 lol.

evgeni:
It does:

```
>>> parser.add_option("--download-method", dest="download_method", default="http", help="Method to download katello-ca-consumer package (e.g. http or https)", metavar="DOWNLOADMETHOD", choices=['http', 'https'])
>>> parser.parse_args([])
(<Values at 0x12daa70: {'download_method': 'http'}>, [])
>>> parser.parse_args(['--download-method', 'https'])
(<Values at 0x137f908: {'download_method': 'https'}>, [])
>>> parser.parse_args(['--download-method', 'htt'])
usage:  [options]

: error: option --download-method: invalid choice: 'htt' (choose from 'http', 'https')
```

bootstrap.py
```diff
@@ -171,8 +173,21 @@ def get_bootstrap_rpm():
         print_generic("To override this behavior, run the script with the --force option. Exiting.")
         sys.exit(1)
 
-    print_generic("Retrieving Client CA Certificate RPMs")
-    exec_failexit("rpm -Uvh http://%s/pub/katello-ca-consumer-latest.noarch.rpm" % options.foreman_fqdn)
+    if options.download_method == "https":
+        print_generic("Writing custom cURL configuration to allow download via HTTPS without certificate verification")
+        curl_config_dir = tempfile.mkdtemp()
+        curl_config = open(("%s/.curlrc" % curl_config_dir),'wb')
```
evgeni: this could (should?) be `os.path.join(curl_config_dir, '.curlrc')` ;)

bootstrap.py
```diff
@@ -162,6 +163,7 @@ def install_prereqs():
 def get_bootstrap_rpm():
     """
     Retrieve Client CA Certificate RPMs from the Satellite 6 server.
+    Uses --insecure options to curl(1) to download via HTTPS
```
evgeni: s/to download/if instructed to download/ ?

bootstrap.py
sideangleside: How dare we have readable code. 👍

bootstrap.py
sideangleside: 👍

PR #159, opened 2017-02-16
modular skip feature

evgeni:
this allows a more modular and granular skipping of certain steps


sideangleside:
Do we want to allow skipping the removal of the old packages (covered today via the `--remove-obsolete-packages` switch)? 

I dont know why you'd _would_, but since we are letting the user skip other previously mandatory steps, it would be consistent.

evgeni:
Right. Will add that.

PR #164, opened 2017-02-22
Allow the user to provide an arbitrary hostname, overridding what is detected via socket.fqdn

sideangleside:
Many users have either hostnames that are short (`hostname -f` or python's `socket.getfqdn` returns a hostname that isn't an FQDN) or non-RFC compliant (containing a character such as an underscore `-` which fails Foreman's hostname validation. 

However, due to $REASONS, the user cannot update his/her system. This PR adds a new option `--override-fqdn` which allows the user to specify the FQDN that will be reported to Foreman

**Prereqs** 

The user needs to set to **False** the `create_new_host_when_facts_are_uploaded` and ` create_new_host_when_reports_are_uploaded` options. This can be done with hammer. 

~~~
hammer settings set --name  create_new_host_when_facts_are_uploaded --value false
hammer settings set --name  create_new_host_when_reports_are_uploaded --value false
~~~


bootstrap.py
```diff
@@ -837,6 +849,7 @@ class MEOptions:
         install_prereqs()
         check_migration_version()
         get_bootstrap_rpm()
+        generate_katello_facts()
```
evgeni, line 897: The call to `generate_katello_facts` in `install_prepreqs` can be removed then?

bootstrap.py
```diff
@@ -285,11 +285,11 @@ def clean_environment():
 
 
 def generate_katello_facts():
-    if FQDN.find(".") != -1 and not os.path.exists('/etc/rhsm/facts/katello.facts'):
-        print_generic("Workaround for FQDN")
-        katellofacts = open('/etc/rhsm/facts/katello.facts', 'w')
-        katellofacts.write('{"network.hostname":"%s"}\n' % (FQDN))
-        katellofacts.close()
+    """Write katello_facts file based on FQDN"""
```
evgeni: Maybe add "should be run *after* katello-ca-consumer RPM was installed to override the generated facts there"?

bootstrap.py
```diff
@@ -285,11 +285,11 @@ def clean_environment():
...
+    print_generic("Writing FQDN katello-fact")
+    katellofacts = open('/etc/rhsm/facts/katello.facts', 'w')
+    katellofacts.write('{"network.hostname-override":"%s"}\n' % (FQDN))
```
evgeni: What is the difference?

bootstrap.py
```diff
@@ -747,11 +738,31 @@ class MEOptions:
         print "\nExample usage: ./bootstrap.py -l admin -s foreman.example.com -o 'Default Organization' -L 'Default Location' -g My_Hostgroup -a My_Activation_Key"
         sys.exit(1)
 
+    # > Gather FQDN, HOSTNAME and DOMAIN using socket.getfqdn()
+    # > If socket.fqdn() returns an FQDN, derive HOSTNAME & DOMAIN using FQDN
+    # > else, HOSTNAME isn't an FQDN
+    # > if user passes --override-fqdn set FQDN, HOSTNAME and DOMAIN to the parameter that is given.
+    FQDN = socket.getfqdn()
```
evgeni: I think `--override-fqdn` should have `default=socket.getfqdn()` and the both codepaths should be merged to one?

bootstrap.py
```diff
@@ -747,11 +738,31 @@ class MEOptions:
...
+    if FQDN.find(".") != -1:
+        HOSTNAME = FQDN.split('.')[0]
+        DOMAIN = FQDN[FQDN.index('.') + 1:]
+    else:
+        HOSTNAME = FQDN
+        DOMAIN = None
+
+    if options.override_fqdn:
+        HOSTNAME = options.override_fqdn
```
evgeni: Missing a `.split('.')[0]`?

bootstrap.py
```diff
@@ -747,11 +738,31 @@ class MEOptions:
...
+        FQDN = options.override_fqdn
+        DOMAIN = FQDN[FQDN.index('.') + 1:]
```
evgeni: this will crash if the user passed a non-fqdn to --override-fqdn (and they will)

evgeni (changes requested):

bootstrap.py
sideangleside: From @jlsherrill: "katello will prefer using network.fqdn,  network.hostname-override, and finally (if none of those are present) network.hostname" I changed our to using network.hostname-override from network.hostname (which we used previously) to be consistent with the katello-ca-consumer RPM.

bootstrap.py
sideangleside: Correct.  	b3fb7f8 removes it.

bootstrap.py
sideangleside: Ack.  added in d68157d

bootstrap.py
sideangleside: This is obsoleted by the refactoring in 15020bf

bootstrap.py
sideangleside: This is obsoleted by the refactoring in 15020bf

sideangleside:
Thanks for the review @evgeni. 

In refactoring the code, I decided to rename `--override-fqdn` to `--fqdn`

bootstrap.py
```diff
@@ -714,6 +707,7 @@ class MEOptions:
     parser.add_option("-l", "--login", dest="login", default='admin', help="Login user for API Calls", metavar="LOGIN")
     parser.add_option("-p", "--password", dest="password", help="Password for specified user. Will prompt if omitted", metavar="PASSWORD")
     parser.add_option("--legacy-login", dest="legacy_login", default='admin', help="Login user for Satellite 5 API Calls", metavar="LOGIN")
+    parser.add_option("--fqdn", dest="fqdn", help="Set an explicit FQDN, overriding detected FQDN from socket.getfqdn()", metavar="OVERRIDE_FQDN", default=socket.getfqdn())
```
evgeni: how about adding `%default` to the help string, thus allowing the user to know what `socket.getfqdn()` detected beforehand? (see https://docs.python.org/2/library/optparse.html#generating-help)

evgeni: I think the metavar should also be renamed to `FQDN` now, no?

evgeni:
Besides the two small comments about the param, LGTM. Did not run the code though.

bootstrap.py
```diff
@@ -714,6 +707,7 @@ class MEOptions:
     parser.add_option("-l", "--login", dest="login", default='admin', help="Login user for API Calls", metavar="LOGIN")
     parser.add_option("-p", "--password", dest="password", help="Password for specified user. Will prompt if omitted", metavar="PASSWORD")
     parser.add_option("--legacy-login", dest="legacy_login", default='admin', help="Login user for Satellite 5 API Calls", metavar="LOGIN")
+    parser.add_option("--fqdn", dest="fqdn", help="Set an explicit FQDN, overriding detected FQDN from socket.getfqdn(), currently detected as %s" % socket.getfqdn(), metavar="FQDN", default=socket.getfqdn())
```
evgeni: The `default` value is stored as `%default`, no need to call `getfqdn()` again.

```python
parser.add_option("--fqdn", dest="fqdn", help="Set an explicit FQDN, overriding detected FQDN from socket.getfqdn(), currently detected as %default", metavar="FQDN", default=socket.getfqdn())```

File: bootstrap.py [hunk 9]
```diff
@@ -714,6 +707,7 @@ class MEOptions:
     parser.add_option("-l", "--login", dest="login", default='admin', help="Login user for API Calls", metavar="LOGIN")
     parser.add_option("-p", "--password", dest="password", help="Password for specified user. Will prompt if omitted", metavar="PASSWORD")
     parser.add_option("--legacy-login", dest="legacy_login", default='admin', help="Login user for Satellite 5 API Calls", metavar="LOGIN")
+    parser.add_option("--fqdn", dest="fqdn", help="Set an explicit FQDN, overriding detected FQDN from socket.getfqdn(), currently detected as %default", metavar="FQDN", default=socket.getfqdn())
```
evgeni: I think this should not be bewteen `--legacy-login` and `--legacy-password` but before?

evgeni:
Needs trivial rebase against master.
Andy maybe add what you have in the "pre-reqs" above to the README? Oh and update the readme to have the newest switches :)

Issue #146, opened 2017-01-04
Satellite complains about blank IP during host entry creation


Tontonitch:
Hello,

I'm trying your script to add existing servers into our Satellite (6.2.5).
I have an issue, as the script fails at the host entry creation as Satellite complains about blank ip.
Please note that we use static IPs for servers, so no dhcp.
```
[...]
[RUNNING], [2017-01-04 17:20:21], [Calling Foreman API to create a host entry associated with the group & org]
url: https://xxxxx:443/api/v2/hosts/
method: POST
data: {
  "host": {
    "managed": "true",
    "name": "server1",
    "hostgroup_id": 8,
    "organization_id": 1,
    "mac": "xx:xx:xx:xx:xx:xx",
    "architecture_id": 1,
    "location_id": 2,
    "domain_id": 1
  }
}
An error occured: HTTP Error 422: Unprocessable Entity
url: https://xxxxx6:443/api/v2/hosts/
code: 422
data: {
  "host": {
    "managed": "true",
    "name": "server1",
    "hostgroup_id": 8,
    "organization_id": 1,
    "mac": "xx:xx:xx:xx:xx:xx",
    "architecture_id": 1,
    "location_id": 2,
    "domain_id": 1
  }
}
error: {
  "error": {
    "errors": {
      "interfaces.ip": [
        "can't be blank"
      ]
    },
    "id": null,
    "full_messages": [
      "Ip can't be blank"
    ]
  }
}
```
BR,
Yannick

Tontonitch:
tested forcing jsondata with the ip setting
`    538     jsondata['host']['ip'] = 'xxx.xxx.xxx.xxx'`

it works

evgeni:
Hey,

what is the subnet definition of the default subnet for hostgroup with id 8?

Tontonitch:
```
[root@xxxx~]# hammer hostgroup info --id=8
Id:                     8
Name:                   HG_RHEL7-SOE-PROD
Title:                  HG_RHEL7-SOE/HG_RHEL7-SOE-PROD
[...]
Parent Id:              3
[root@xxxx~]#
```
```
[root@xxxx~]# hammer hostgroup info --id=3
Id:                     3
Name:                   HG_RHEL7-SOE
Title:                  HG_RHEL7-SOE
Operating System:       RedHat 7.3
[...]
Subnet:                 DC subnet
[...]
[root@xxxx~]# 
```
```
[...]

PR #165, opened 2017-02-24
option to specify the ip of the server

evgeni:
Also gathers the primary IP address if none was given

### DIFF
--- a/README.md
+++ b/README.md
@@ -399,7 +399,7 @@ Options:
   --skip=SKIP           Skip the listed steps (choices: ['foreman', 'puppet',
                         'migration', 'prereq-update', 'katello-agent',
                         'remove-obsolete-packages'])
-
+  --ip=IP               IP address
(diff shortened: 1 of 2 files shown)


evgeni:
This should fix #146

sideangleside (changes requested):
What's a good usage for the --ip switch? Generally when we add a new option, I'd like the README to show the user its usage, especially since it is a non-mandatory option. 

Is it simply useful where the host has multiple interfaces and you wish to control which is used for contacting Foreman? Or are there other usages?

evgeni:
it is only used in the foreman host object, e.g. to contact the machine from foreman (rex). clarified that (at least I hope I did)

sideangleside (approved):

sideangleside:
I tested on a couple of systems.  Doc text looks good. Merging. :+1:

PR #168, opened 2017-03-06
Add capsule migration option

wzzrd:
I added a --new-capsule option, as discussed on Rocket.Chat. 

It took some going back and forth and trying out various existing methods. I used some, like clean_katello_agent(), and ignored some others (clean_puppet()) for simplicity's sake.

I have tested with a RHEL7 machine, but I see no reason this shouldn't work on a RHEL6 client. The only piece of code that is not mine that I'm potentially hitting is the put_json() function, which seems to be called in disassociate_host(). Haven't tested, but I don't think it breaks anything.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,9 @@ bootstrap Script for migrating existing running systems to Foreman with the Kate
 This script can take a system that is registered to Spacewalk, Satellite 5, Red Hat
 Network Classic and get it registered to Foreman & Katello.
 
+Optionally, you can also move systems from one Satellite 6 system or Capsule to
+another by using the --new-capsule option.
+
 # What does the Script do?
 
 * Identify which systems management platform is the system registered to (Classic/Sat5 or None)  then perform the following
@@ -28,6 +31,17 @@ Network Classic and get it registered to Foreman & Katello.
 * Running subscription-manager (with the user provided activation key) to register the system.
 * Configuring the system with a proper Puppet configuration pointing at Foreman
 * Removing/disabling old RHN Classic packages/daemons (rhnsd, osad, etc)
+
+## System already registered to a Satellite 6 server / Capsule (--new-capsule)
+* Clean the existing Katello agent installation
+* Install the Katello consumer RPM for the target Satellite 6 server / Capsule
+* Install the Katello agent software again, using the configuration for the
+  target Satellite 6 / Capsule server
+* Make API calls to update the Puppet master, Puppet CA, content source and
+  OpenSCAP proxy IDs
+* Re-enable rhsmcertd
+* Update the Puppet configuration for the system to point to the 
+* Restart Puppet and call for the user to go sign the CSR
 
 # Assumptions
 
@@ -169,6 +183,18 @@ By default, bootstrap.py does not delete the system's profile from the legacy pl
 ### Migrating a system from one Foreman + Katello installation to another.
 
 There are times where it is necessary to migrate clients from one Foreman + Katello installation to another. For instance, in lieu of upgrading an older Foreman + Katello installation, you choose to build a new installation in parallel. bootstrap.py can then be used to migrate clients from one Foreman + Katello installation to another. Simply provide the `--force` option, and bootstrap.py will remove the previous `katello-ca-consumer-*` package (from the old system), and will install the `katello-ca-consumer-*` package (from the new system), and continue registration as usual.
+
+### Migrating a system from one Satellite 6 / Capsule to another in the same infrastructure
+
+In order to manually balance the load over multiple Capsule servers, you might
+want to move some existing systems to newly deployed Capsules. You can easily
+do this by running the bootstrap.py script like so:
+
+
+~~~
+# ./bootstrap.py --new-capsule \
+  --server new.capsule.server
+~~~
 
 ### Enabling additional repositories at registration time.
 
@@ -418,6 +444,8 @@ Options:
   --ip=IP               IPv4 address of the primary interface in Foreman
                         (defaults to the address used to make request to
                         Foreman)
+  --new-capsule         Migrate the system to a new Capsule server, pass the
+                        name of the new Capsule with the --server option
 ~~~
 
 # For developers:
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -160,13 +160,13 @@ def install_prereqs():
         yum("update", "yum openssl python")
 
 
-def get_bootstrap_rpm():
+def get_bootstrap_rpm(clean=False):
     """
     Retrieve Client CA Certificate RPMs from the Satellite 6 server.
     Uses --insecure options to curl(1) if instructed to download via HTTPS
     If called with --force, calls clean_katello_agent().
     """
-    if options.force:
+    if options.force or clean == True:
         clean_katello_agent()
     if os.path.exists('/etc/rhsm/ca/katello-server-ca.pem'):
         print_generic("A Katello CA certificate is already installed. Assuming system is registered")
@@ -936,8 +936,7 @@ if __name__ == '__main__':
         # > Puppet, OpenSCAP and update Puppet configuration
         # > MANUAL SIGNING OF CSR OR MANUALLY CREATING AUTO-SIGN RULE STILL REQUIRED!
         # > API doesn't have a public provision for creating auto-sign entries yet!
-        options.force = True  # Need to set this to make get_bootstrap_rpm() to perform the right way
-        get_bootstrap_rpm()
+        get_bootstrap_rpm(clean=True)
         install_katello_agent()
         API_PORT = get_api_port()
 
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -936,9 +936,8 @@ if __name__ == '__main__':
         # > Puppet, OpenSCAP and update Puppet configuration
         # > MANUAL SIGNING OF CSR OR MANUALLY CREATING AUTO-SIGN RULE STILL REQUIRED!
         # > API doesn't have a public provision for creating auto-sign entries yet!
-        print_running("Removing katello-*, gofer and installing new Katello certificate")
-        clean_katello_agent()
-        exec_failexit("rpm -Uvh http://%s/pub/katello-ca-consumer-latest.noarch.rpm" % options.foreman_fqdn)
+        options.force = True  # Need to set this to make get_bootstrap_rpm() to perform the right way
+        get_bootstrap_rpm()
         install_katello_agent()
         API_PORT = get_api_port()
 
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -166,7 +166,7 @@ def get_bootstrap_rpm(clean=False):
     Uses --insecure options to curl(1) if instructed to download via HTTPS
     If called with --force, calls clean_katello_agent().
     """
-    if options.force or clean == True:
+    if clean == True:
         clean_katello_agent()
     if os.path.exists('/etc/rhsm/ca/katello-server-ca.pem'):
         print_generic("A Katello CA certificate is already installed. Assuming system is registered")
@@ -919,7 +919,7 @@ if __name__ == '__main__':
         print_generic('This system is registered to RHN. Attempting to migrate via rhn-classic-migrate-to-rhsm')
         install_prereqs()
         check_migration_version()
-        get_bootstrap_rpm()
+        get_bootstrap_rpm(clean=options.force)
         generate_katello_facts()
         API_PORT = get_api_port()
         if 'foreman' not in options.skip:
@@ -971,7 +971,7 @@ if __name__ == '__main__':
         # > ELSE get CA RPM, optionally create host,
         # >      register via subscription-manager
         print_generic('This system is not registered to RHN. Attempting to register via subscription-manager')
-        get_bootstrap_rpm()
+        get_bootstrap_rpm(clean=options.force)
         generate_katello_facts()
         API_PORT = get_api_port()
         if 'foreman' not in options.skip:
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -894,8 +894,8 @@ if __name__ == '__main__':
     # > Clean the environment from LD_... variables
     clean_environment()
 
-    # > IF RHEL 5 and not removing, prepare the migration.
-    if not options.remove and int(RELEASE[0]) == 5:
+    # > IF RHEL 5, not removing, and not moving to new capsule prepare the migration.
+    if not options.remove and int(RELEASE[0]) == 5 and not options.new_capsule:
         prepare_rhel5_migration()
 
     if options.remove:
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -459,9 +459,9 @@ def delete_json(url):
     return call_api(url, method='DELETE')
 
 
-def put_json(url):
+def put_json(url, jdata=None):
     """Use `call_api` to place a "PUT" REST API call."""
-    return call_api(url, method='PUT')
+    return call_api(url, data=jdata, method='PUT')
 
 
 def return_matching_foreman_key(api_name, search_key, return_key, null_result_ok=False):
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -955,11 +955,11 @@ if __name__ == '__main__':
         print_running("Stopping the Puppet agent for configuration update")
         exec_failok("/sbin/service puppet stop")   # failok because people might be running Puppet from cron
 
+        # Not using clean_puppet() and install_puppet_agent() here, because
+        # that would nuke custom /etc/puppet/puppet.conf files, which might
+        # yield undesirable results.
         print_running("Updating Puppet configuration")
-        pupcfg = SafeConfigParser()
-        pupcfg.read("/etc/puppet/puppet.conf")
-        pupcfg.set("agent", "server", options.foreman_fqdn)
-        pupcfg.write("/etc/puppet/puppet.conf")
+        exec_failexit("sed -i 's/^[[:space:]]*server.*/   server     = %s/' /etc/puppet/puppet.conf" % options.foreman_fqdn)
         delete_directory("/var/lib/puppet/ssl")
         delete_file("/var/lib/puppet/client_data/catalog/%s.json" % FQDN)
 
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -164,7 +164,8 @@ def get_bootstrap_rpm(clean=False):
     """
     Retrieve Client CA Certificate RPMs from the Satellite 6 server.
     Uses --insecure options to curl(1) if instructed to download via HTTPS
-    If called with --force, calls clean_katello_agent().
+    This function is usually called with clean=options.force, which ensures
+    clean_katello_agent() is called is --force is specified.
     """
     if clean:
         clean_katello_agent()
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -463,10 +463,12 @@ def put_json(url, jdata=None):
     """Use `call_api` to place a "PUT" REST API call."""
     return call_api(url, data=jdata, method='PUT')
 
+
 def update_host_capsule_mapping(server, port, attribute, capsule_id, host_id):
     url = "https://" + server + ":" + str(port) + "/api/v2/hosts/" + str(host_id)
     jdata = json.loads('{"host": {"%s": "%s"}}' % (attribute, capsule_id))
     return put_json(url, jdata)
+
 
 def return_matching_foreman_key(api_name, search_key, return_key, null_result_ok=False):
     """
@@ -890,7 +892,6 @@ if __name__ == '__main__':
             except ImportError:
                 print_error("Could not install python-simplejson")
 
-
     # > Clean the environment from LD_... variables
     clean_environment()
 
@@ -930,7 +931,7 @@ if __name__ == '__main__':
             enable_repos()
     elif options.new_capsule and options.foreman_fqdn:
         # > ELIF new_capsule and foreman_fqdn set, will migrate to other capsule
-        # 
+        #
         # > will replace CA certificate, reinstall katello-agent, gofer
         # > update system definition to point to new capsule for content,
         # > Puppet, OpenSCAP and update Puppet configuration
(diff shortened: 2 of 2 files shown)


bootstrap.py
```diff
@@ -879,7 +891,6 @@ class MEOptions:
                 import simplejson as json
             except ImportError:
                 print_error("Could not install python-simplejson")
-                sys.exit(1)
```
evgeni: Why? simplejson is required on RHEL5.

bootstrap.py
```diff
@@ -918,6 +929,44 @@ class MEOptions:
...
+
+        print_running("Restarting goferd and rhsmcertd")
+        exec_failexit("/sbin/service rhsmcertd restart")
+        exec_failexit("/sbin/service goferd restart")
+
+        print_running("Stopping the Puppet agent for configuration update")
+        exec_failok("/sbin/service puppet stop")   # failok because people might be running Puppet from cron
+
+        print_running("Updating Puppet configuration")
+        exec_failexit("/usr/bin/puppet config set --section agent server %s" % options.foreman_fqdn)
```
evgeni: this does not work on RHEL5, the puppet agent there does not support "puppet config set" :(

bootstrap.py
```diff
@@ -918,6 +929,44 @@ class MEOptions:
...
+        # > ELIF new_capsule and foreman_fqdn set, will migrate to other capsule
+        #
+        # > will replace CA certificate, reinstall katello-agent, gofer
+        # > update system definition to point to new capsule for content,
+        # > Puppet, OpenSCAP and update Puppet configuration
+        # > MANUAL SIGNING OF CSR OR MANUALLY CREATING AUTO-SIGN RULE STILL REQUIRED!
+        # > API doesn't have a public provision for creating auto-sign entries yet!
+        print_running("Removing katello-*, gofer and installing new Katello certificate")
+        clean_katello_agent()
+        exec_failexit("rpm -Uvh http://%s/pub/katello-ca-consumer-latest.noarch.rpm" % options.foreman_fqdn)
```
evgeni: This should be `get_bootstrap_rpm()`?

bootstrap.py
```diff
@@ -918,6 +929,44 @@ class MEOptions:
...
+        capsule_id = return_matching_katello_key('capsules', 'name="%s"' % options.foreman_fqdn, 'id', False)
+        host_id = return_matching_foreman_key('hosts', 'name="%s"' % FQDN, 'id', False)
+
+        update_host_capsule_mapping(options.foreman_fqdn, API_PORT, "puppet_proxy_id", capsule_id, host_id)
+        update_host_capsule_mapping(options.foreman_fqdn, API_PORT, "puppet_ca_proxy_id", capsule_id, host_id)
+        update_host_capsule_mapping(options.foreman_fqdn, API_PORT, "content_source_id", capsule_id, host_id)
+        update_host_capsule_mapping(options.foreman_fqdn, API_PORT, "openscap_proxy_id", capsule_id, host_id)
+
+        print_running("Restarting goferd and rhsmcertd")
+        exec_failexit("/sbin/service rhsmcertd restart")
```
evgeni: This should be `enable_rhsmcertd()`?

bootstrap.py
```diff
@@ -918,6 +929,44 @@ class MEOptions:
...
+        exec_failexit("/sbin/service goferd restart")
```
evgeni: This is already done in `install_katello_agent()`, why repeating it?

bootstrap.py
```diff
@@ -918,6 +929,44 @@ class MEOptions:
...
+
+        print_running("Stopping the Puppet agent for configuration update")
+        exec_failok("/sbin/service puppet stop")   # failok because people might be running Puppet from cron
+
+        print_running("Updating Puppet configuration")
```
evgeni: I wonder if this all should be just

```
clean_puppet()
install_puppet_agent()
```

bootstrap.py
```diff
@@ -765,6 +772,8 @@ class MEOptions:
         options.skip.append('puppet')
     if not options.removepkgs:
         options.skip.append('remove-obsolete-packages')
+    if options.new_capsule:
```
evgeni: This should be done after `get_bootstrap_rpm()` via `API_PORT = get_api_port()`. Also this makes moving *to* the internal capsules impossible.

evgeni (changes requested):

bootstrap.py
wzzrd: No. That nukes puppet.conf, which might be bad for some users.

bootstrap.py
wzzrd: I didn't read your code well enough :) Will change, new pull request to follow

bootstrap.py
wzzrd: Will change, new pull request to follow

bootstrap.py
wzzrd: Quite. This seems to be a remnant of my own refactoring. Will change, new pull request to follow

bootstrap.py
wzzrd: My bad. Removed this by accident :|

bootstrap.py
wzzrd: Does it show really badly that I don't do this for a living ;) Will change, new pull request to follow

bootstrap.py
wzzrd: @evgeni Ok, so if `puppet conf` is out of the question, next best thing probably is to run over `/etc/puppet/puppet.conf` and replace the `server =` line with a new one. I don't want to go all regexpy, if I can prevent it :)

Again, using `install_puppet_agent()` will nuke an existing `/etc/puppet/puppet.conf` and I'm note sure if that's a good idea. People might have customized that for good reasons.

Thoughts?

bootstrap.py
wzzrd: Let me look into using ConfigParser a bit. Might fit this nicely. Again: don't want to nuke people's `/etc/puppet/puppet.conf` just to change a single settings :D

wzzrd: It turns out SafeConfigParser is pretty picky on what it accepts as a valid config file. This is a no-go.  I might end up running `sed`instead, ffs :P

evgeni: Yeah, I tried ConfigParser and it choked pretty badly on the default puppet.conf, so that is why we have this ugly blob to begin with.

bootstrap.py
evgeni: Yupp. It would be at least consistent with a fresh bootstrap, but I understand your concern.

bootstrap.py
wzzrd: Ok, went with sed for now. Two actually, because stock RHEL5 puppet.conf has a ca_server *and* a server option.

wzzrd:
Updated pull request cf. comments by @evgeni. Eagerly awaiting more :)

bootstrap.py
```diff
@@ -918,6 +928,46 @@ class MEOptions:
...
             enable_repos()
+    elif options.new_capsule and options.foreman_fqdn:
+        # > ELIF new_capsule and foreman_fqdn set, will migrate to other capsule
+        #
+        # > will replace CA certificate, reinstall katello-agent, gofer
+        # > update system definition to point to new capsule for content,
+        # > Puppet, OpenSCAP and update Puppet configuration
+        # > MANUAL SIGNING OF CSR OR MANUALLY CREATING AUTO-SIGN RULE STILL REQUIRED!
+        # > API doesn't have a public provision for creating auto-sign entries yet!
+        options.force = True  # Need to set this to make get_bootstrap_rpm() to perform the right way
```
evgeni: can we change `get_bootstrap_rpm` to accept a `clean` Boolean, default it to `False` but pass `options.force` every time we call it instead. And then here don't set force to `true` but call `get_bootstrap_rpm(clean=True)`?

Just want to make sure that in future we don't actually want to omit something on `force=False` when moving capsules and can't or forget about it.

bootstrap.py
```diff
@@ -918,6 +928,46 @@ class MEOptions:
...
+
+        print_running("Calling Foreman API to update content source, puppet master, puppet ca and openscap proxy records for %s" % FQDN)
+        capsule_id = return_matching_katello_key('capsules', 'name="%s"' % options.foreman_fqdn, 'id', False)
+        host_id = return_matching_foreman_key('hosts', 'name="%s"' % FQDN, 'id', False)
+        update_host_capsule_mapping(options.foreman_fqdn, API_PORT, "puppet_proxy_id", capsule_id, host_id)
+        update_host_capsule_mapping(options.foreman_fqdn, API_PORT, "puppet_ca_proxy_id", capsule_id, host_id)
+        update_host_capsule_mapping(options.foreman_fqdn, API_PORT, "content_source_id", capsule_id, host_id)
+        update_host_capsule_mapping(options.foreman_fqdn, API_PORT, "openscap_proxy_id", capsule_id, host_id)
+
+        print_running("Restarting rhsmcertd")
```
evgeni: I think this should be "Enabling and restarting rhsmcertd" and be moved *inside* `enable_rhsmcertd` so it gets printed on the other parts too.

bootstrap.py
```diff
@@ -459,9 +459,15 @@ def delete_json(url):
...
 
 
-def put_json(url):
+def put_json(url, jdata=None):
     """Use `call_api` to place a "PUT" REST API call."""
-    return call_api(url, method='PUT')
+    return call_api(url, data=jdata, method='PUT')
+
+
+def update_host_capsule_mapping(server, port, attribute, capsule_id, host_id):
```
evgeni: We usually don't have `server` and `port` in the other functions, and just default to the global stuff. While not the nicest way, it would be consistent. I am torn.

evgeni (changes requested):

evgeni:
I wonder if "new capsule" is the right wording here.

1. we call it "Foreman Proxy" in Foreman/Katello
2. this can also be used to move the client to the embedded proxy, which the current wording "forbids"?

wzzrd:
@evgeni regarding the last comment: they're called Capsules throughout the script right now. I agree proxy would probably be more upstream-friendly, but then we need to be consistent with this change as well and replace all occurences of 'capsule' with 'smart proxy' (which might be confusing for downstream users).

bootstrap.py
wzzrd: Can live with that. Am updating some code now, will check whether this works later this week and also include the switch back to built-in proxy as test case.

bootstrap.py
wzzrd: So we could do this in two ways:

1. make get_bootstrap_rpm run clean_katello_agent() when either options.force is True *or* clean is True
2. change the code to always call get_bootstrap_rpm() like get_bootstrap_rpm(clean=options.force)

Preferences?

evgeni:
@wzzrd agreed, as the olde code uses "Capsule", let's keep it Capsule for now.

bootstrap.py
evgeni: My initial thoughts were along the lines of 2, as it is less code.

@sideangleside what about you?

wzzrd: Went with option 2 for now

bootstrap.py
wzzrd: Makes sense. Done.

bootstrap.py
```diff
@@ -160,13 +160,13 @@ def install_prereqs():
         yum("update", "yum openssl python")
 
 
-def get_bootstrap_rpm():
+def get_bootstrap_rpm(clean=False):
     """
     Retrieve Client CA Certificate RPMs from the Satellite 6 server.
     Uses --insecure options to curl(1) if instructed to download via HTTPS
     If called with --force, calls clean_katello_agent().
```
evgeni: this comment should be updated to match the new `clean` param

bootstrap.py
```diff
@@ -918,11 +929,49 @@ class MEOptions:
...
+        # > update system definition to point to new capsule for content,
+        # > Puppet, OpenSCAP and update Puppet configuration
+        # > MANUAL SIGNING OF CSR OR MANUALLY CREATING AUTO-SIGN RULE STILL REQUIRED!
+        # > API doesn't have a public provision for creating auto-sign entries yet!
+        get_bootstrap_rpm(clean=True)
+        install_katello_agent()
+        API_PORT = get_api_port()
+
+        print_running("Calling Foreman API to update content source, puppet master, puppet ca and openscap proxy records for %s" % FQDN)
+        capsule_id = return_matching_katello_key('capsules', 'name="%s"' % options.foreman_fqdn, 'id', False)
```
evgeni: any reason not to hit the Foreman API for that?

```
capsule_id = return_matching_foreman_key('smart_proxies', 'name="%s"' % options.foreman_fqdn, 'id', False)
```

evgeni:
Two very nit-picky comments, but overall :+1:  from a code/style perspective.
Would love to run some local tests on that, but not sure I will find the time in the next week(s) :(

evgeni (approved):

bootstrap.py
wzzrd: Will add.

bootstrap.py
wzzrd: Not really (I think). Just picked the first one in the API docs that fitted my requirements. Would have to test whether or not the Foreman API works as well (but would guess so). Any reason as why this would matter?

evgeni: This would make (this part of the code) also work on Foreman installs without Katello. Plus I'd hope we can someday get rid of the Katello-specific API completely.

I gave the APIs on 1.15 upstream nightly and 6.2.8 Satellite a shot yesteday, the results looked identical enough for your needs.

wzzrd:
Sounds good. I'll /katello/foreman/ then

README.md
```diff
@@ -170,6 +184,18 @@ By default, bootstrap.py does not delete the system's profile from the legacy pl
...
 
+### Migrating a system from one Satellite 6 / Capsule to another in the same infrastructure
+
+In order to manually balance the load over multiple Capsule servers, you might
+want to move some existing systems to newly deployed Capsules. You can easily
+do this by running the bootstrap.py script like so:
+
+
+~~~
+# ./bootstrap.py --new-capsule \
```
evgeni, line 195: this needs credentials, no?

evgeni:
One nitpick for the README.md: we usually don't say "Satellite 6" but "Foreman + Katello", but we do name the products in the "migrating from X" section. So not sure what I would actually change in your doc.

README.md
```diff
@@ -7,6 +7,9 @@ bootstrap Script for migrating existing running systems to Foreman with the Kate
 This script can take a system that is registered to Spacewalk, Satellite 5, Red Hat
 Network Classic and get it registered to Foreman & Katello.
 
+Optionally, you can also move systems from one Satellite 6 system or Capsule to
+another by using the --new-capsule option.
```
evgeni, line 11: `--new-capsule` so that it properly formated

bootstrap.py
```diff
@@ -918,11 +930,49 @@ class MEOptions:
...
+        capsule_id = return_matching_foreman_key('smart_proxies', 'name="%s"' % options.foreman_fqdn, 'id', False)
+        host_id = return_matching_foreman_key('hosts', 'name="%s"' % FQDN, 'id', False)
+        update_host_capsule_mapping("puppet_proxy_id", capsule_id, host_id)
+        update_host_capsule_mapping("puppet_ca_proxy_id", capsule_id, host_id)
+        update_host_capsule_mapping("content_source_id", capsule_id, host_id)
+        update_host_capsule_mapping("openscap_proxy_id", capsule_id, host_id)
+
+        enable_rhsmcertd()
+
+        print_running("Stopping the Puppet agent for configuration update")
```
evgeni, line 955: this should be wrapped in `if 'puppet' not in options.skip`?

bootstrap.py
```diff
@@ -918,11 +930,49 @@ class MEOptions:
...
+        # > will replace CA certificate, reinstall katello-agent, gofer
+        # > update system definition to point to new capsule for content,
+        # > Puppet, OpenSCAP and update Puppet configuration
+        # > MANUAL SIGNING OF CSR OR MANUALLY CREATING AUTO-SIGN RULE STILL REQUIRED!
+        # > API doesn't have a public provision for creating auto-sign entries yet!
+        get_bootstrap_rpm(clean=True)
+        install_katello_agent()
+        API_PORT = get_api_port()
+
+        print_running("Calling Foreman API to update content source, puppet master, puppet ca and openscap proxy records for %s" % FQDN)
```
evgeni, line 945: this should be wrapped in `if 'foreman' not in options.skip`?

wzzrd:
Let me update / improve that readme in a separate pull request. I give the
whole thing a makeover

On Fri, Mar 24, 2017, 10:25 Evgeni Golov <notifications@github.com> wrote:

One nitpick for the README.md: we usually don't say "Satellite 6" but
"Foreman + Katello", but we do name the products in the "migrating from X"
section. So not sure what I would actually change in your doc.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/Katello/katello-client-bootstrap/pull/168#issuecomment-288973083>,
or mute the thread

evgeni (approved):

bootstrap.py
```diff
@@ -918,11 +930,49 @@ class MEOptions:
...
+        capsule_id = return_matching_foreman_key('smart_proxies', 'name="%s"' % options.foreman_fqdn, 'id', False)
+        host_id = return_matching_foreman_key('hosts', 'name="%s"' % FQDN, 'id', False)
+        update_host_capsule_mapping("puppet_proxy_id", capsule_id, host_id)
+        update_host_capsule_mapping("puppet_ca_proxy_id", capsule_id, host_id)
+        update_host_capsule_mapping("content_source_id", capsule_id, host_id)
+        update_host_capsule_mapping("openscap_proxy_id", capsule_id, host_id)
```
sideangleside, line 951: It is not guaranteed that every capsule runs the OpenSCAP feature. In testing, migrating from one Capsule to another (where the destination did not have the OpenSCAP feature), bootstrap.py fails with an error message similar to

~~~
RUNNING], [2017-03-30 06:22:12], [Calling Foreman API to update content source, puppet master, puppet ca and openscap proxy records for devnode-115.example.com] 
An error occured: HTTP Error 422: Unprocessable Entity
url: https://capsule-2.example.com:8443/api/v2/hosts/231
code: 422
data: {
  "host": {
    "openscap_proxy_id": "3"
  }
}
error: {
  "error": {
    "errors": {
      "openscap_proxy_id": [
        "must have Openscap feature"
      ]
    }, 
    "id": 231, 
    "full_messages": [
      "Openscap Capsule must have Openscap feature"
    ]
  }
}

~~~

It may make sense to check the Capsule's features, and then update the host if (and only if) it has the OpenSCAP feature.

bootstrap.py
```diff
@@ -918,11 +930,49 @@ class MEOptions:
...
+        # that would nuke custom /etc/puppet/puppet.conf files, which might
+        # yield undesirable results.
+        print_running("Updating Puppet configuration")
+        exec_failexit("sed -i 's/^[[:space:]]*server.*/   server     = %s/' /etc/puppet/puppet.conf" % options.foreman_fqdn)
+        exec_failok("sed -i 's/^[[:space:]]*ca_server.*/   server     = %s/' /etc/puppet/puppet.conf" % options.foreman_fqdn)  # For RHEL5 stock puppet.conf
+        delete_directory("/var/lib/puppet/ssl")
+        delete_file("/var/lib/puppet/client_data/catalog/%s.json" % FQDN)
+
+        print_running("Starting Puppet and waiting for cert; please go and sign it now!")
+        exec_failok("/usr/bin/puppet agent --noop -t --waitforcert 15")
```
sideangleside, line 968: For consistency would it make sense to update this line to match what is in the `install_puppet_agent()` function: 

~~~
    print_generic("Running Puppet in noop mode to generate SSL certs")
~~~

sideangleside (changes requested):

sideangleside:
Also, is the expectation that the user is allowed to switch hostgroups while switching capsules? Depending on how they have their hostgroups setup, they may want to.

bootstrap.py
wzzrd: Fair point, will look into that.

bootstrap.py
wzzrd: Dito

wzzrd:
So about the hostgroup thing, I see your point. This will take some time though as it adds a new layer of functionality to this feature. Alternatively, changing the hostgroup could be a version 2.0 and we now add docs to manipulate that part through the UI, where changing a hostgroup is trivial.

evgeni:
@wzzrd what's the status of this?

bootstrap.py
```diff
@@ -459,9 +461,15 @@ def delete_json(url):
...
-def put_json(url):
+def put_json(url, jdata=None):
     """Use `call_api` to place a "PUT" REST API call."""
-    return call_api(url, method='PUT')
+    return call_api(url, data=jdata, method='PUT')
+
+
+def update_host_capsule_mapping(attribute, capsule_id, host_id):
+    url = "https://" + options.foreman_fqdn + ":" + str(API_PORT) + "/api/v2/hosts/" + str(host_id)
+    jdata = json.loads('{"host": {"%s": "%s"}}' % (attribute, capsule_id))
```
ekohl, line 471: I believe this line is equal to ```jdata = {'host': {attribute: capsule_id}}```

bootstrap.py
wzzrd: The Puppet stuff only, right? Not the content source? What about the OpenSCAP proxy?

Should I wrap the puppet_proxy_id and puppet_ca_proxy_id lines in `if 'puppet' not in options.skip`, too?

bootstrap.py
wzzrd: OK

wzzrd:
Meh. Finally took some time to work on this again. 

Some issues I could need some help on: 

- @sideangleside I can check whether a proxy supports OpenSCAP, but the v2 API doesn't seem to expose openscap_proxy_id in the host object. So though I can *set* the openscap_proxy_id, I cannot query whether or not a host currently has an OpenSCAP proxy configured. 

I might be missing something: it's been a while since I started this and it's late, but if I'm missing that the API exposes this somewhere else, pointers appreciated. 

- @evgeni For now, I've wrapped the update_host_capsule_mapping lines as follows, please advise:
```
        if 'foreman' in options.skip or 'puppet' in options.skip:
            update_host_capsule_mapping("puppet_proxy_id", capsule_id, host_id)
            update_host_capsule_mapping("puppet_ca_proxy_id", capsule_id, host_id)
        if 'foreman' in options.skip:
            update_host_capsule_mapping("openscap_proxy_id", capsule_id, host_id) 
        update_host_capsule_mapping("content_source_id", capsule_id, host_id) 
```

wzzrd:
I'm going to file a new PR tomorrow and close this one. Too many changes compared to last time I pushed this.

PR #170, opened 2017-03-15
Update bootstrap.py for check rpm

peppos:
Add a function for check if on the server is installed the Satellite or Capsule product.
If katello-capsule rpm is installed exit because Satellite or Capsule product is installed.

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -638,19 +638,21 @@ def get_api_port():
 print "Foreman Bootstrap Script"
 print "This script is designed to register new systems or to migrate an existing system to a Foreman server with Katello"
 
-def check_prerequisite():
-   rpm_sat = ['satellite','satellite-capsule']
-   check_rpm_installed(rpm_sat)
 
+def check_prerequisite():
+    rpm_sat = ['satellite','satellite-capsule']
+    check_rpm_installed(rpm_sat)
+
+    
 def check_rpm_installed(rpm_sat):
     ts = rpm.TransactionSet()
     headers = ts.dbMatch()
-
     for h in headers:
         if h['name'] in rpm_sat:
             print_error("RPM Satellite/capsule installed")
             sys.exit(1)
 
+            
 def prepare_rhel5_migration():
     """
     Execute specific preparations steps for RHEL 5. Older releases of RHEL 5
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -638,10 +638,12 @@ def get_api_port():
 print "Foreman Bootstrap Script"
 print "This script is designed to register new systems or to migrate an existing system to a Foreman server with Katello"
 
+
 def check_prerequisite():
     rpm_sat = "katello-capsule"
     check_rpm_installed(rpm_sat)
 
+    
 def check_rpm_installed(rpm_sat):
     ts = rpm.TransactionSet()
     headers = ts.dbMatch('name', rpm_sat)
@@ -656,6 +658,7 @@ def check_rpm_installed(rpm_sat):
         print_error("RPM Satellite/capsule installed")
         sys.exit(1)
 
+        
 def prepare_rhel5_migration():
     """
     Execute specific preparations steps for RHEL 5. Older releases of RHEL 5
@@ -708,7 +711,7 @@ if __name__ == '__main__':
     
     # Check if Satellite or Capsule are installed or not
     check_prerequisite()
-
+    
     # > Register our better HTTP processor as default opener for URLs.
     opener = urllib2.build_opener(BetterHTTPErrorProcessor)
     urllib2.install_opener(opener)
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -643,7 +643,7 @@ def check_prerequisite():
     rpm_sat = "katello-capsule"
     check_rpm_installed(rpm_sat)
 
-    
+	
 def check_rpm_installed(rpm_sat):
     ts = rpm.TransactionSet()
     headers = ts.dbMatch('name', rpm_sat)
@@ -658,7 +658,7 @@ def check_rpm_installed(rpm_sat):
         print_error("RPM Satellite/capsule installed")
         sys.exit(1)
 
-        
+
 def prepare_rhel5_migration():
     """
     Execute specific preparations steps for RHEL 5. Older releases of RHEL 5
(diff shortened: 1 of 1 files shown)


bootstrap.py
```diff
@@ -688,7 +708,10 @@ class MEOptions:
 
 
 if __name__ == '__main__':
-
+    
+    # Check if Satellite or Capsule are installed or not
+    check_prerequisite()
```
sideangleside: Might want to put this after options parsing. (basically after [here](https://github.com/Katello/katello-client-bootstrap/blob/master/bootstrap.py#L866). No need in checking until we have a sane set of options.  Also, you get an error when running with no options. 

~~~
./bootstrap.py 
Foreman Bootstrap Script
This script is designed to register new systems or to migrate an existing system to a Foreman server with Katello
[ERROR], [2017-03-15 10:52:59], EXITING: [RPM Satellite/capsule installed] failed to execute properly.
Must specify server, login, organization, hostgroup and activation key.  See usage:
Usage: bootstrap.py [options]


~~~

bootstrap.py
```diff
@@ -639,6 +639,26 @@ def get_api_port():
 print "This script is designed to register new systems or to migrate an existing system to a Foreman server with Katello"
 
 
+def check_prerequisite():
+    rpm_sat = "katello-capsule"
```
sideangleside: the `katello-capsule` package doesn't exist on Katello servers, just Smart Proxies that have content (f.k.a Capsules)

bootstrap.py
```diff
@@ -639,6 +639,26 @@ def get_api_port():
...
+    check_rpm_installed(rpm_sat)
+
+	
+def check_rpm_installed(rpm_sat):
+    ts = rpm.TransactionSet()
+    headers = ts.dbMatch('name', rpm_sat)
```
sideangleside: It may make sense to check against a list of packages, not just a single package, as depending on configuration (and upstream vs. downstream) the list of packages that define Katello or Satellite may be > 1. This could also keep from repeating `headers = ts.dbMatch('name', rpm_sat)` for each package. 
 
Example:
~~~
    rpm_sat = ['katello-capsule','satellite']
    check_rpm_installed(rpm_sat)

def check_rpm_installed(rpm_sat):
    ts = rpm.TransactionSet()
    headers = ts.dbMatch()

    for h in headers:
        if h['name'] in rpm_sat:
            print_error("RPM Satellite/capsule installed")
            sys.exit(1)

~~~

bootstrap.py
Rocco83: any suggestion on package name to be used?

sideangleside: Sure do. Upstream you have `katello` and `katello-capsule` (pre 3.3) (/me forgot that Katello 3.2 and lower used `katello-capsule`). Katello > 3.3 you'd want to check for `foreman-proxy-content`. Downstream you'd want to check for `satellite`and `satellite-capsule`.  Though in fairness, if you run `bootstrap.py` on a system functioning as a Smart Proxy with content or a Capsule, it is already registered and `bootstrap.py` will fail because the system already has the CA certificate. But it doesn't hurt to be much more explicit.

evgeni: Given that the upstream packages will be installed downstream too, I don't think we have to explicitly check for the downstream names, no?

sideangleside: Yes, I forgot about that. (I am so used to `rpm -q satellite` downstream) +1 to what @evgeni says. Just check the upstream names. We can always patch if it changes.

sideangleside:
@peppos status on this PR?

bootstrap.py
```diff
@@ -639,6 +639,20 @@ def get_api_port():
 print "This script is designed to register new systems or to migrate an existing system to a Foreman server with Katello"
 
 
+def check_prerequisite():
+    rpm_sat = ['satellite', 'satellite-capsule']
```
evgeni: I think this should be

    ['katello', 'foreman-proxy-content', 'katello-capsule']

that will work both up- and downstream, and on new (3.4) and old (3.2) releases.

bootstrap.py
```diff
@@ -689,6 +703,9 @@ class MEOptions:
 
 if __name__ == '__main__':
 
+    # Check if Satellite or Capsule are installed or not
+    check_prerequisite()
```
evgeni: This should probably go somewhere around https://github.com/peppos/katello-client-bootstrap/blob/6198337c0b1172d90d238d0b009e3fb9883212de/bootstrap.py#L880

(either before or after the `root` check).

evgeni (changes requested):

bootstrap.py
sideangleside: Would it make sense to in the PR to also add the older spacewalk packages (like spacewalk-backend) to prevent users from running bootstrap.py against a Spacewalk/Satellite Server/Proxy. There are reports of users (see https://bugzilla.redhat.com/show_bug.cgi?id=1448722) doing this and breaking their Satellite 5 Server/Proxy (as we remove a ton of rhn* packages at the end of the script).

evgeni: Yeah, why not.

peppos: Yes for me, but I not able to test it. I need a specewalk :-(
You have a rpm name?

bootstrap.py
peppos: I have moved the check after the root check.

bootstrap.py
sideangleside: spacewalk-proxy-common & spacewalk-backend for Spacewalk/Sat5 proxy & Server respectively.

peppos: Thanks @sideangleside!

bootstrap.py
```diff
@@ -639,6 +639,20 @@ def get_api_port():
...
+    rpm_sat = ['katello', 'foreman-proxy-content', 'katello-capsule', 'spacewalk-proxy-common', 'spacewalk-backend']
+    check_rpm_installed(rpm_sat)
+
+
+def check_rpm_installed(rpm_sat):
+    ts = rpm.TransactionSet()
+    headers = ts.dbMatch()
+    for h in headers:
+        if h['name'] in rpm_sat:
+            print_error("RPM Satellite/capsule installed")
```
evgeni, line 652: How about:

```python
print_error("%s RPM found. bootstrap.py should not be used on a Katello/Spacewalk/Satellite host." % (h['name']))
```

bootstrap.py
```diff
@@ -639,6 +639,20 @@ def get_api_port():
 print "This script is designed to register new systems or to migrate an existing system to a Foreman server with Katello"
 
 
+def check_prerequisite():
+    rpm_sat = ['katello', 'foreman-proxy-content', 'katello-capsule', 'spacewalk-proxy-common', 'spacewalk-backend']
```
evgeni, line 643: I think this list can go inside `check_rpm_installed` and then we can drop `check_prerequisite` as a function completelly?

evgeni (approved):

evgeni:
@peppos thanks! I added comments, but none of them should block a merge, IMHO.

PR #177, opened 2017-03-29
Add puppet-enable --skip option. Update README

### LINKED ISSUE #174: [RFE] - Allow user to disable autostart of Puppet Agent
From [RHBZ1435972](https://bugzilla.redhat.com/show_bug.cgi?id=1435972) - As a user, I'd like to disable the puppet agent from starting by default (via chkconfig and the service commands). This allows me to deploy puppet but keep it disabled until needed.

sideangleside:
Fixes #174

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -316,7 +316,6 @@ def install_puppet_agent():
     puppet_env = return_puppetenv_for_hg(return_matching_foreman_key('hostgroups', 'title="%s"' % options.hostgroup, 'id', False))
     print_generic("Installing the Puppet Agent")
     yum("install", "puppet")
-    exec_failexit("/sbin/chkconfig puppet on")
     puppet_conf = open('/etc/puppet/puppet.conf', 'wb')
     puppet_conf.write("""
 [main]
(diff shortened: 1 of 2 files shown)


bootstrap.py
```diff
@@ -339,7 +338,9 @@ def install_puppet_agent():
     print_generic("Visit the UI and approve this certificate via Infrastructure->Capsules")
     print_generic("if auto-signing is disabled")
     exec_failexit("/usr/bin/puppet agent --test --noop --tags no_such_tag --waitforcert 10")
-    exec_failexit("/sbin/service puppet restart")
+    if 'puppet-enable' not in options.skip:
+        exec_failexit("/sbin/service puppet restart")
```
evgeni: I'd have sorted those two the other way round, first enable, then restart, but it does not strictly hurt.

evgeni (approved):

bootstrap.py
sideangleside: Consistent behavior is a good thing, as our enabling/starting of rhsmcertd and goferd is enable first, (re) start second.

sideangleside:
Flipped the options around in 0e9184a. Squashed the two commits.

PR #178, opened 2017-03-30
Add --deps-repository-* options to allow user to specify repo with subscription-manager + dependencies

### LINKED ISSUE #172: Add support for installing subscription-manager from the http kickstart repositories
For clients who do not have access to subscription-manager via a Spacewalk or Satellite 5 server, we should automatically attempt to hook into the freely available http based Kickstart repositories on the Katello/Satellite 6 server if synchronized.

This will streamline the bootstrap process and ensure subscription-manager is properly installed.

sideangleside:
Fixes #172

Major changes

- Added two new options
~~~
--deps-repository-url" - URL to a repository that contains 
       the subscription-manager RPMs
--deps-repository-gpg-key" - GPG Key to the repository that contains 
      the subscription-manager RPMs. 
      (defaults to file:///etc/pki/rpm-gpg/RPM-GPG-KEY-redhat-release")
~~~

- Call `install_prereqs()` even in non-migration use-cases
- update `install_prereqs()` to check if `subscription-manager` is installed or available. If not installed, attempt to install the package if it is available. Otherwise, exit. (and inform user of the switches above). 

Example usage

~~~
./bootstrap.py -s server.example.com   \
  -o Example \
  -a ak_Reg_To_Dev_EL7 \
  -g 'RHEL7/BS-Dev'   \
  -L Philadelphia  \
  --deps-repository-url "http://server.example.com/pulp/repos/Example/Library/content/dist/rhel/server/7/7.2/x86_64/kickstart/"
~~~

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -7,6 +7,7 @@ Use `pydoc ./bootstrap.py` to get the documentation.
 Use `awk -F'# >' 'NF>1 {print $2}' ./bootstrap.py` to see the flow.
 """
 
+import yum
 import getpass
 import urllib2
 import base64
@@ -124,9 +125,15 @@ def delete_directory(directoryname):
         sys.exit(1)
 
 
-def yum(command, pkgs=""):
-    """Helper function to call a yum command on a list of packages."""
-    exec_failexit("/usr/bin/yum -y %s %s" % (command, pkgs))
+def call_yum(command, params="", failonerror=True):
+    """
+    Helper function to call a yum command on a list of packages.
+    pass failonerror = False to make yum commands non-fatal
+    """
+    if failonerror:
+        exec_failexit("/usr/bin/yum -y %s %s" % (command, params))
+    else:
+        exec_failok("/usr/bin/yum -y %s %s" % (command, params))
 
 
 def check_migration_version():
@@ -149,15 +156,52 @@ def check_migration_version():
         sys.exit(1)
 
 
+def setup_yum_repo(url, gpg_key):
+    """
+    Configures a yum repository at /etc/yum.repos.d/katello-client-bootstrap-deps.repo
+    """
+    submanrepoconfig = SafeConfigParser()
+    submanrepoconfig.add_section('kt-bootstrap')
+    submanrepoconfig.set('kt-bootstrap', 'name', 'Subscription-manager dependencies for katello-client-bootstrap')
+    submanrepoconfig.set('kt-bootstrap', 'baseurl', url)
+    submanrepoconfig.set('kt-bootstrap', 'enabled', '1')
+    submanrepoconfig.set('kt-bootstrap', 'gpgcheck', '1')
+    submanrepoconfig.set('kt-bootstrap', 'gpgkey', gpg_key)
+    submanrepoconfig.write(open('/etc/yum.repos.d/katello-client-bootstrap-deps.repo', 'w'))
+    print_generic('Building yum metadata cache. This may take a few minutes')
+    call_yum('makecache')
+
+
 def install_prereqs():
     """
     Install subscription manager and its prerequisites.
+    If subscription-manager is installed already, check to see if we are migrating. If yes, install subscription-manager-migration.
+    Else if subscription-manager and subscription-manager-migration are available in a configured repo, install them both.
+    Otherwise, exit and inform user that we cannot proceed
     """
-    print_generic("Installing subscription manager prerequisites")
-    yum("remove", "subscription-manager-gnome")
-    yum("install", "subscription-manager 'subscription-manager-migration-*'")
+    print_generic("Checking subscription manager prerequisites")
+    if options.deps_repository_url:
+        print_generic("Enabling %s as a repository for dependency RPMs" % options.deps_repository_url)
+        setup_yum_repo(options.deps_repository_url, options.deps_repository_gpg_key)
+    yb = yum.YumBase()
+    pkg_list = yb.doPackageLists(patterns=['subscription-manager'])
+    call_yum("remove", "subscription-manager-gnome")
+    if pkg_list.installed:
+        if check_rhn_registration() and 'migration' not in options.skip:
+            print_generic("installing subscription-manager-migration")
+            call_yum("install", "'subscription-manager-migration-*'", False)
+        print_generic("subscription-manager is installed already. Attempting update")
+        call_yum("update", "subscription-manager 'subscription-manager-migration-*'", False)
+    elif pkg_list.available:
+        print_generic("subscription-manager NOT installed. Installing.")
+        call_yum("install", "subscription-manager 'subscription-manager-migration-*'")
+    else:
+        print_error("Cannot find subscription-manager in any configured repository. Consider using the --deps-repository-url switch to specify a repository with the subscription-manager RPMs")
+        sys.exit(1)
     if 'prereq-update' not in options.skip:
-        yum("update", "yum openssl python")
+        call_yum("update", "yum openssl python", False)
+    if options.deps_repository_url:
+        delete_file('/etc/yum.repos.d/katello-client-bootstrap-deps.repo')
 
 
 def get_bootstrap_rpm():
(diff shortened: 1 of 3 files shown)


bootstrap.py
```diff
@@ -7,6 +7,7 @@
 Use `awk -F'# >' 'NF>1 {print $2}' ./bootstrap.py` to see the flow.
 """
 
+import yum as yumApi
```
evgeni: yuk. can we instead rename our `yum()` function to `call_yum()` or something? I *hate* renaming imports.

bootstrap.py
```diff
@@ -149,15 +150,46 @@ def check_migration_version():
...
+    submanrepoconfig = SafeConfigParser()
+    submanrepoconfig.add_section('kt-bootstrap')
+    submanrepoconfig.set('kt-bootstrap', 'name', 'Subscription-manager dependencies for katello-client-bootstrap')
+    submanrepoconfig.set('kt-bootstrap', 'baseurl', url)
+    submanrepoconfig.set('kt-bootstrap', 'enabled', '1')
+    submanrepoconfig.set('kt-bootstrap', 'gpgcheck', '1')
+    submanrepoconfig.set('kt-bootstrap', 'gpgkey', gpg_key)
+    submanrepoconfig.write(open('/etc/yum.repos.d/katello-client-bootstrap-deps.repo', 'w'))
+    print_generic('Building yum metadata cache. This may take a few minutes')
+    exec_failexit('/usr/bin/yum makecache')
```
evgeni: can't this be `call_yum("makecache")`?

bootstrap.py
```diff
@@ -149,15 +150,46 @@ def check_migration_version():
...
+    if options.deps_repository_url:
+        print_generic("Enabling %s as a repository for dependency RPMs" % options.deps_repository_url)
+        setup_yum_repo(options.deps_repository_url, options.deps_repository_gpg_key)
+    yb = yumApi.YumBase()
+    yb.conf.cache = os.geteuid() != 1
+    pkg_list = yb.doPackageLists(patterns=['subscription-manager'])
     yum("remove", "subscription-manager-gnome")
-    yum("install", "subscription-manager 'subscription-manager-migration-*'")
+    if pkg_list.installed:
+        print_generic("subscription-manager is installed already")
```
evgeni: shall we try upgrading it here?

bootstrap.py
```diff
@@ -149,15 +150,46 @@ def check_migration_version():
...
     """
     Install subscription manager and its prerequisites.
     """
-    print_generic("Installing subscription manager prerequisites")
+    print_generic("Checking subscription manager prerequisites")
+    if options.deps_repository_url:
+        print_generic("Enabling %s as a repository for dependency RPMs" % options.deps_repository_url)
+        setup_yum_repo(options.deps_repository_url, options.deps_repository_gpg_key)
+    yb = yumApi.YumBase()
+    yb.conf.cache = os.geteuid() != 1
```
evgeni: UID 1? Why that? root has 0, and we don't allow running the script w/o root.

bootstrap.py
evgeni: also we are not installing migration rpm anymore if subman is installed?

evgeni (changes requested):

bootstrap.py
sideangleside: I was debating on which I hated more:

- renaming our `yum` function.  (Which means a LOT of code gets touched)
- renaming the import. 

They both suck.

bootstrap.py
sideangleside: Yep +1

bootstrap.py
sideangleside: True.

bootstrap.py
sideangleside: Possibly. What is yum's behavior in the scenario where the user had no repos configured and I ran a `yum install subscription-manager`. Does it fail because it can't install or update the package or does it gracefully return because the package _is_ installed already

sideangleside: Actually we need it for both registration and migration cases. (As for older RHEL5 we need to grab product certs). Also, if we choose to add better support for checking for product certs before we register, we'd need it.

sideangleside: In my testing yum barfs here. If we try to upgrade sub-man but no updates are available. yum fails.

sideangleside:
- Changed the `yum()` function to `call_yum()`
- Added a `warnonerror` option to the `call_yum()` function. 
- Attempt non-fatal updates of subscription-manager, python and openssl.

bootstrap.py
```diff
@@ -124,9 +125,15 @@ def delete_directory(directoryname):
         sys.exit(1)
 
 
-def yum(command, pkgs=""):
-    """Helper function to call a yum command on a list of packages."""
-    exec_failexit("/usr/bin/yum -y %s %s" % (command, pkgs))
+def call_yum(command, pkgs="", warnonerror=False):
```
evgeni: I would have called the param `failonerror` and default it to `True` as the name `warnonerror` somehow implies that otherwise nothing happens on error.
Also, while we edit this function already, can we rename `pkgs` to `params` or something? when we call not install/remove, the pkgs name does not really match.

evgeni (approved):

evgeni:
How crazy would you call me, if I would propose to add the following:

```diff
--- bootstrap-dep-orig.py	2017-04-05 12:54:37.044000000 +0000
+++ bootstrap-dep.py	2017-04-05 13:01:20.379000000 +0000
@@ -926,6 +926,15 @@
                 print_error("Could not install python-simplejson")
                 sys.exit(1)
 
+    if 'foreman' in options.skip:
+        org_label = options.org
+    else:
+        org_label = return_matching_katello_key('organizations', 'name="%s"' % options.org, 'label', False)
+
+    if options.deps_repository_url == 'auto':
+        options.deps_repository_url = 'http://%s/pulp/repos/%s/Library/content/dist/rhel/server/%s/%s/%s/kickstart/' % (options.foreman_fqdn, org_label, RELEASE[0], RELEASE, ARCHITECTURE)
+
+
     # > Clean the environment from LD_... variables
     clean_environment()
 
```

bootstrap.py
```diff
@@ -149,15 +156,46 @@ def check_migration_version():
...
+    print_generic("Checking subscription manager prerequisites")
+    if options.deps_repository_url:
+        print_generic("Enabling %s as a repository for dependency RPMs" % options.deps_repository_url)
+        setup_yum_repo(options.deps_repository_url, options.deps_repository_gpg_key)
+    yb = yum.YumBase()
+    pkg_list = yb.doPackageLists(patterns=['subscription-manager'])
+    call_yum("remove", "subscription-manager-gnome")
+    if pkg_list.installed:
+        print_generic("subscription-manager is installed already. Attempting update")
+        call_yum("update", "subscription-manager 'subscription-manager-migration-*'", True)
```
evgeni: in the case sub-man is installed, but sub-man-migration is not, this will not install sub-man-migration :(

evgeni (changes requested):

evgeni:
No, scratch that auto thing. That would work for RHEL, but not for CentOS or Fedora. So we'd need to have more cases... Ugh.

bootstrap.py
sideangleside: +1

bootstrap.py
sideangleside: True, but it gets more interesting. In migration (from RHN/Sat5) use-cases on RHEL6/7, we need `subscription-manager-migration` for the `rhn-migrate-classic-to-rhsm` command.  For RHEL5 we dont as it doesn't support activation keys.  We call `install_prereqs` in both registration & migration use cases, there is (currently) no way (unless I add a param to this function) to discern between them. So failing in a registration use-case due to lack of `subscription-manager-migration` would be wrong.

bergsjoh:
Tested with --deps-repository-url=DEPS_REPOSITORY_URL.
LGTM

sideangleside:
Added some code to ensure that `subscription-manager-migration` is installed during migration use cases. @evgeni feel like testing?

evgeni:
@sideangleside looks good. are you going to fix the comment in https://github.com/Katello/katello-client-bootstrap/pull/178/files#r109910077 too?

evgeni (approved):

evgeni:
ACK. squash and merge away :)

PR #182, opened 2017-05-24
fall back to port 443 if for some reason it is not in rhsm.conf

evgeni:
this prevents errors like:

    ConfigParser.NoOptionError: No option 'port' in section: 'server'

This can happen when you install katello-ca-consumer on a machine with
old subscription-manager, which triggers the "alternative" configuration
using sed, the old rhsm.conf did not have a port entry for whatever
reason and sed thus will not add the missing line.

See https://github.com/Katello/puppet-certs/blob/master/templates/rhsm-katello-reconfigure.erb#L63


sideangleside:
+1 LGTM:

PR #187, opened 2017-06-08
add --install-packages switch. Update README

### LINKED ISSUE #183: [RFE] Allow users to install arbitrary packages via bootstrap.py
As a user, I'd like to have the ability to install arbitrary packages when I bootstrap a system. Something like a `--install-packages` switch to install a comma separated list of RPMs. All it would offer is the install. (Think of it as a courtesy for folks without configuration management / automation.)

sideangleside:
Fixes #183


evgeni (approved):
I would have called the option `additional-packages`, but no hard feelings here. ACK.

README.md
```diff
@@ -388,7 +404,7 @@ Options:
                         Password for specified user. Will prompt if omitted
   --fqdn=FQDN           Set an explicit FQDN, overriding detected FQDN from
                         socket.getfqdn(), currently detected as
-                        client.example.com
+                        bs-1234.example.org
```
evgeni: I think the "currently detected" part should not be in the readme, no? :)

sideangleside: Yeah, I saw that after sending the PR last night. Airplane wifi was crap so I couldn't amend.

PR #189, opened 2017-06-19
Allow skipping enable/start of goferd

evgeni:
See http://projects.theforeman.org/issues/19393 for details.

@sideangleside @jlsherrill I'd like your reviews ;)

I think we could (and should) drop `katello-agent` from the skip list now, as goferd is really what we want to skip.

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -320,11 +320,12 @@ def clean_katello_agent():
 
 
 def install_katello_agent():
-    """Install Katello agent (aka Gofer) and activate /start it."""
+    """Install Katello agent (aka Gofer) and maybe activate/start it."""
     print_generic("Installing the Katello agent")
     call_yum("install", "katello-agent")
-    exec_failexit("/sbin/chkconfig goferd on")
-    exec_failexit("/sbin/service goferd restart")
+    if 'goferd' not in options.skip:
+        exec_failexit("/sbin/chkconfig goferd on")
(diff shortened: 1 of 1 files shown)


sideangleside:
> I think we could (and should) drop katello-agent from the skip list now, as goferd is really what we want to skip.

I think we'd need both. I can imagine that there are scenarios where 

- the user doesnt wants to install `katello-agent` 
- the user does want to install `katello-agent`, but not enable the `goferd` daemon. (Think Satellite 6.2 users who want to use the new goferless work. They'd need to explicitly **disable** goferd.)
- the user does want to install `katello-agent`, but explicitly wants to enable the `goferd` daemon.  (My understanding is that newer `katello-agent` packages will default to `goferd` being disabled)

evgeni:
In which scenario would one not want to install `katello-agent` at all?

sideangleside:
Users who don't care about errata applicability or users who are using Foreman or Satellite as just an ENC for puppet.

evgeni:
mkay.

rplevka:
isn't goferd enabled && started automatically on install for all versions of katello agent? I can see this is quite old PR so I guess the enable-start part already lost its sense.

beav:
@evgeni needs rebase (or closing if it's no longer needed)

evgeni:
closing as obsolete, will be handled better via the fix for #251

PR #196, opened 2017-07-05
Fixes filter_string function

PizarroG:
./bootstrap.py -s sat6 -l admin -p pass -a ak --skip=foreman -o org --force

Foreman Bootstrap Script
This script is designed to register new systems or to migrate an existing system to a Foreman server with Katello
[NOTIFICATION], [2017-07-05 22:42:09], [This system is not registered to RHN. Attempting to register via subscription-manager] 
[NOTIFICATION], [2017-07-05 22:42:09], [Checking subscription manager prerequisites] 
Loaded plugins: langpacks, product-id
  File "./bootstrap.py", line 1014, in <module>
    install_prereqs()
  File "./bootstrap.py", line 200, in install_prereqs
    call_yum("remove", "subscription-manager-gnome")
  File "./bootstrap.py", line 146, in call_yum
    exec_failexit("/usr/bin/yum -y %s %s" % (command, params))
  File "./bootstrap.py", line 98, in exec_failexit
    filtered_command = filter_string(command)
  File "./bootstrap.py", line 54, in filter_string
    return string.replace(options.password, '******').replace(options.legacy_password, '******')
TypeError: expected a character buffer object

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -51,7 +51,11 @@ error_colors = {
 
 def filter_string(string):
     """Helper function to filter out passwords from strings"""
-    return string.replace(str(options.password), '******').replace(str(options.legacy_password), '******')
+    if options.password:
+        string = string.replace(options.password, '******')
+    if options.legacy_password:
+        string = string.replace(options.legacy_password, '******')
+    return string
 
 
 def print_error(msg):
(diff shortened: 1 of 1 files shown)


evgeni:
Good catch!
What do you think about the following fix instead?
```python
if options.password:
    string = string.replace(options.password, '******')
if options.legacy_password:
    string = string.replace(options.legacy_password, '******')
return string
```

that way we don't replace stray "None"s in the output ;)

PizarroG:
Sounds good to me :)

evgeni:
Thanks!

PR #197, opened 2017-07-18
add ansible example

evgeni:


dglinder:
Since running the bootstrap.py can take quite a while, I'd suggest the addition of a "stat" module call and only run the bootstrap.py if the client does NOT have a .pem file.  The bootstrap.py uses this file as a signal to exit but doing this in Ansible eliminates the Python interpreter call altogether.  (And the "force_all_plays_to_run" can be set to true to override this check if necessary.)

```
 - name: Check for existing Satellite certificate file
   stat: path=/etc/rhsm/ca/katello-server-ca.pem
   register: katello_pem

 - name: Execute the bootstrap.py script
   command: "/var/tmp/bootstrap.py ...."
   when: not katello_pem.stat.exists or force_all_plays_to_run|bool
```

evgeni:
I must admit I don't see the benefit here. Checking that pem is pretty much the first thing bootstrap does.
And unless you pass `bootstrap_additional_args: --force` to the playbook, you get exactly the behavior you want, no? The only thing you would save is the start up of the python interpreter itself. And that's pretty quick.

dglinder:
Was the PEM check always there?  I though an early version I used re-ran it regardless.  Adding in the check saves copying the script.  Again, not a big deal in the grand scheme of things so in your environment it might make sense to copy and run regardless.

evgeni:
The PEM check is there since https://github.com/Katello/katello-client-bootstrap/pull/120, so no, not always :)

dglinder:
Ok - I retract my comment/addition as it is not directly aligned with the usage of the bootstrap.py script within an Ansible playbook.

sean797:
+1 on having a general example. Though it probably should include pulling bootstrap.py from `https://<capsule fqdn>/pub/bootstrap.py`

evgeni:
it does pull by default?

sean797:
Can this example also delete `bootstrap.py` after?

dglinder:
@sean797 - Sure, the example would just need a "file:" module call with "state: absent".  Something like this (untested):
```
 - name: Execute the bootstrap.py script
   file:
     path: /var/tmp/bootstrap.py
     state: absent
```

evgeni:
this has so far only got positive feedback, and given it's not a code change, but an example, I am going to just merge that now :)

Issue #142, opened 2016-12-08
systemctl / chkconfig


bergsjoh:
In install_katello_agent and install_puppet_agent bootstrap is using chkconfig.

In some cases (where you might run automation, with non-full environments etc) you will end up with errors due to chkconfig redirection to systemctl on RHEL7 doesn't work.

Proposed fix: Run systemctl commands for rhel7, chkconfig for < rhel7.

evgeni:
In which environment would that not work?

Quick check on a EL7 machine yields:
```
[root@localhost ~]# chkconfig postfix on
Note: Forwarding request to 'systemctl enable postfix.service'.
Created symlink from /etc/systemd/system/multi-user.target.wants/postfix.service to /usr/lib/systemd/system/postfix.service.
[root@localhost ~]# chkconfig postfix off
Note: Forwarding request to 'systemctl disable postfix.service'.
Removed symlink /etc/systemd/system/multi-user.target.wants/postfix.service.
```

So yeah, we could add an if/else to the right places, to make the code clearer. But it should work already today.

bergsjoh:
We encountered issues with this due to crappy automation tools. I worked around it, was basically a suggestion for cleanup, not a bug in bootstrap, but rather my tools around it.

alosadagrande:
+1. Same error due to automation tool.

PR #203, opened 2017-08-11
Add condition to run and enable goferd,puppet,rhsmcertd via systemctl

alosadagrande:
Add condition to execute and enable: goferd, puppet and rhsmcertd via systemctl for RHEL >= 7 instead of chkconfig and service. This fixes some errors running bootstrap.py on RHEL7 using third party automation systems (issue #142)

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -264,13 +264,8 @@ def enable_rhsmcertd():
     """
     Enable and restart the rhsmcertd service
     """
-
-    if int(RELEASE[0]) >= 7:
-        exec_failexit("/usr/bin/systemctl enable rhsmcertd")
-        exec_failexit("/usr/bin/systemctl restart rhsmcertd")
-    else:
-        exec_failexit("/sbin/chkconfig rhsmcertd on")
-        exec_failexit("/sbin/service rhsmcertd restart")
+    enable_service("rhsmcertd")
+    start_service("rhsmcertd")
 
 
 def migrate_systems(org_name, activationkey):
@@ -342,13 +337,8 @@ def install_katello_agent():
     """Install Katello agent (aka Gofer) and activate /start it."""
     print_generic("Installing the Katello agent")
     call_yum("install", "katello-agent")
-
-    if int(RELEASE[0]) >= 7:
-        exec_failexit("/usr/bin/systemctl enable goferd")
-        exec_failexit("/usr/bin/systemctl restart goferd")
-    else:
-        exec_failexit("/sbin/chkconfig goferd on")
-        exec_failexit("/sbin/service goferd restart")
+    enable_service("goferd")
+    start_service("restart", "goferd")
 
 
 def clean_puppet():
@@ -384,11 +374,7 @@ def install_puppet_agent():
     puppet_env = return_puppetenv_for_hg(return_matching_foreman_key('hostgroups', 'title="%s"' % options.hostgroup, 'id', False))
     print_generic("Installing the Puppet Agent")
     call_yum("install", "puppet")
-
-    if int(RELEASE[0]) >= 7:
-        exec_failexit("/usr/bin/systemctl enable puppet")
-    else:
-        exec_failexit("/sbin/chkconfig puppet on")
+    enable_service("puppet")
 
     puppet_conf = open('/etc/puppet/puppet.conf', 'wb')
     puppet_conf.write("""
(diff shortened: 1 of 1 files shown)


evgeni (changes requested):
I'd prefer to have wrapper functions like `enable_service(service)` and `restart_service(service)` that do the `if/else` game and the caller just does not care.

Also checking `int(RELEASE[0])` might work now, but would make addition of other OSes in the future harder. I'd detect systemd by checking that `/run/systemd/` exists.

alosadagrande:
Add both wrappers and check if systemd exists based on path suggested. If that path does not exist I assume /sbin/service is the only choice available to enable and run services. Assumed that since bootstrap.py goal is to take a RHEL or RHEL based client registered to Foreman & Katello

bootstrap.py
```diff
@@ -264,8 +264,8 @@ def enable_rhsmcertd():
     """
     Enable and restart the rhsmcertd service
     """
-    exec_failexit("/sbin/chkconfig rhsmcertd on")
-    exec_failexit("/sbin/service rhsmcertd restart")
+    enable_service("rhsmcertd")
+    start_service("rhsmcertd")
```
evgeni: This needs to be `start_service("rhsmcertd", "restart")`

bootstrap.py
```diff
@@ -337,8 +337,8 @@ def install_katello_agent():
     """Install Katello agent (aka Gofer) and activate /start it."""
     print_generic("Installing the Katello agent")
     call_yum("install", "katello-agent")
-    exec_failexit("/sbin/chkconfig goferd on")
-    exec_failexit("/sbin/service goferd restart")
+    enable_service("goferd")
+    start_service("restart", "goferd")
```
evgeni: This needs to be `start_service("goferd", "restart")`

bootstrap.py
```diff
@@ -401,8 +402,8 @@ def install_puppet_agent():
     print_generic("if auto-signing is disabled")
     exec_failexit("/usr/bin/puppet agent --test --noop --tags no_such_tag --waitforcert 10")
     if 'puppet-enable' not in options.skip:
-        exec_failexit("/sbin/chkconfig puppet on")
-        exec_failexit("/sbin/service puppet restart")
+        enable_service("puppet")
+        start_service("restart", "puppet")
```
evgeni: This needs to be `start_service("puppet", "restart")`

bootstrap.py
```diff
@@ -766,6 +767,40 @@ class MEOptions:
...
+        else:
+            exec_failexit("/sbin/chkconfig %s on" % (service))
+    else:
+        if os.path.exists("/run/systemd"):
+            exec_failok("/usr/bin/systemctl enable %s" % (service))
+        else:
+            exec_failok("/sbin/chkconfig %s on" % (service))
+
+
+def start_service(service, command="start", failonerror=True):
```
evgeni: I like the way you made it possible to start/restart/stop the service. Shall we call the wrapper just `service`, as it can do more than start?

evgeni (changes requested):

evgeni:
The assumption is perfectly fine for now. If we ever have to support more than those, we'll adjust the wrappers and be done :)

bootstrap.py
alosadagrande: It can't be cause the command parameter has assigned a default value equal to start at definition time. That's why the service which has no default value at definition time must be prior to command otherwise you should expect a python error.

bootstrap.py
alosadagrande: It can't be cause the command parameter has assigned a default value equal to start at definition time. That's why the service which has no default value at definition time must be prior to command otherwise you should expect a python error.

bootstrap.py
alosadagrande: OK, I can change the function to allow any command and then avoid the default value of command (command = start) and keep the order suggested. 

I would suggest exec_service for clarity so that we can keep service as a parameter, which is already defined in more functions.

alosadagrande:
I would suggest exec_service for clarity so that we can keep service as a parameter, which is already defined in more functions along the code.

evgeni (approved):

evgeni:
@sideangleside would like an ACK from you, but I think this is good.

sideangleside:
:+1:

PR #205, opened 2017-08-22
Adding --release to subscription-manager

soukron:
Adding --release <release> at registration time is needed when using a different release from default (ie 7Server vs 7.3). Otherwise, subscription is done properly (but using 7Server) and then katello (or other RPMs) installation fail because the CV/repo contains only 7.3 repos. 

Also, note that register_systems function seems to be prepared for this as release is one of the arguments received but never used in the function. This commit just finishes the work.

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -316,6 +316,8 @@ def register_systems(org_name, activationkey, release):
     options.smargs += " --serverurl=https://%s:%s/rhsm --baseurl=https://%s/pulp/repos" % (options.foreman_fqdn, API_PORT, options.foreman_fqdn)
     if options.force:
         options.smargs += " --force"
+    if release:
+        options.smargs += " --release %s" % release
     exec_failexit("/usr/sbin/subscription-manager register --org '%s' --name '%s' --activationkey '%s' %s" % (org_label, FQDN, activationkey, options.smargs))
     enable_rhsmcertd()
 


alosadagrande (commented):
Should not be options.release ?

soukron:
That's what I thought but for some reason is in the function definition so I used it.

sideangleside:
I seem to remember that there was a logical reason why we didn't do this. Note that the `--release` option exists for `subscription-manager`, but not `rhn-classic-migrate-to-rhsm`, and that _might_ be why we didn't implement it. (It would only work in registration use cases, but not migration from RHN/Sat5) 

@evgeni : does this ring a bell?

evgeni:
Nope, as you were going to implement it differently:

https://github.com/Katello/katello-client-bootstrap/commit/b63a0cea7f25b29d7a25f610d680da61b80e0f41

(without saying how ;))

soukron:
@sideangleside @evgeni as said, reading the code, this PR makes all sense as the function definition is ready, the call in line number 1025 passes **options.release** when calling the function and **options.release** is not used in any other place in the code... so... crystal clear for me 😄

evgeni:
how about this: https://github.com/evgeni/katello-client-bootstrap/commit/release?
based on your patch, but also adds `sub-man release --set` in the migration path, and drops the default release value, so that we don't pass 7.x when the user wants 7Server

soukron:
Didnt notice the migration path, I could have added it also.

Regarding the default value, if no arg is specified it doesn't add a release so subscription-manager execution, but dropping the default sounds good also just in case.

evgeni:
With your code I get:

```
[root@ccbc81b3c17a /]# python bootstrap.py -s katello.example.com --skip foreman -a key --force
…
[RUNNING], [2017-08-28 18:28:59], [/usr/sbin/subscription-manager register --org 'Default Organization' --name 'ccbc81b3c17a' --activationkey 'key'  --serverurl=https://katello.example.com:443/rhsm --baseurl=https://katello.example.com/pulp/repos --force --release 7.3.1611] 
```

So it is passing the detected release at the moment.

soukron:
Oh I see.
That's because I check the parameter passed to the function instead of the
argument in the bootstrap execution.

Do you want me to update the PR or do you want to proceed with yours, which
is more complete?

Regards
Sergio

evgeni:
Not exactly, this behavior is because you still have `default=RELEASE` in your code for the `--release` parameter, so it is never unset.

I don't really care which branch gets merged as long as it makes both of us happy ;)
If we pretend my patch is more complete, and it still works as you expect it, it's probably easier to just merge mine.
Can you test my code in your environment?

soukron:
Actually I already tested it when you sent it :D
I used to override the 7Server release set in the AK with 7.2 and also it's
more complete so I'm happy with it if you eventually merge it.

evgeni:
OK, opened https://github.com/Katello/katello-client-bootstrap/pull/211

evgeni:
Closing as #211 was merged now

PR #206, opened 2017-08-24
clean up local subscription data when force-registering

evgeni:
subscription-manager in some versions does not like if there are local
certificates left that are signed by a previous CA. see RHBZ#1484830 for
details.

to workaround this behaviour, let's call sub-man unregister and sub-man
clean to wipe out all local information (but also try to unregister, if
that is possible).

RHBZ#1484830: https://bugzilla.redhat.com/show_bug.cgi?id=1484830


ehelms (approved):

sideangleside:
:+1:

PR #209, opened 2017-08-25
Validate user credendials before trying to bootstrap.

bergsjoh:
Silently validate user credentials before throwing a an exception and print userfriendly errors msgs.

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -851,7 +851,7 @@ def guess_api_port(host):
         port = 443
         sock.close()
         return port
-    else
+    else:
         return None
 
 
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -781,18 +781,25 @@ def prepare_rhel5_migration():
 def validate_login():
     """
     Function to authenticate a user to validate login credentials before
-    running any other API calls. Strip down output so it doesn't print a
+    running any other API calls. Since we don't have API_PORT setup yet
+    we need to make a test. Strip down output so it doesn't print a
     big exception and scare users. This function is going to be used
     before we have the certificates from katello-consumer rpm installed,
     so we need to disable SSL verification since it's now enabled
     per default in RHEL7.4+.
     """
-    myurl = "https://" + options.foreman_fqdn + ":" + API_PORT + "/api/v2/organizations/"
+    port = guess_api_port(options.foreman_fqdn)
+    if not port:
+        print "ERROR, ports 443 and 8443 seems closed, cannot connect."
+        sys.exit(2)
+
+    myurl = "https://" + options.foreman_fqdn + ":" + port + "/api/v2/organizations/"
     try:
         jsonresult=call_api(myurl,no_verify_ssl=True,silent=True)
     except:
         print "ERROR, could not authenticate username. Please try again."
         sys.exit(2)
+
 
 def enable_service(service, failonerror=True):
     """
@@ -827,6 +834,25 @@ def exec_service(service, command, failonerror=True):
             exec_failok("/usr/bin/systemctl %s %s" % (command, service))
         else:
             exec_failok("/sbin/service %s %s" % (service, command))
+
+
+def guess_api_port(host):
+    """
+    We need to guess the port to be used to validate login credentials
+    since we don't have API_PORT globally defined yet.
+    """
+    sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM))
+    sock.settimeout(2)
+    if sock.connect_ex((host,8443)) == 0:
+        port = 8443
+        sock.close()
+        return port
+    else if sock.connect_ex((host,443)) == 0:
+        port = 443
+        sock.close()
+        return port
+    else
+        return None
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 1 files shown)


bootstrap.py
```diff
@@ -766,6 +776,48 @@ class MEOptions:
...
+
+
+def guess_api_port(host):
+    """
+    We need to guess the port to be used to validate login credentials
+    since we don't have API_PORT globally defined yet.
+    """
+    sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    sock.settimeout(2)
+    if sock.connect_ex((host, 8443)) == 0:
```
evgeni: the `if` and `else` parts have identical code, so how about:
```python
for port in [8443, 443]:
    if sock.connect_ex((host, port)) == 0:
        sock.close()
        return str(port)
return None
```

bootstrap.py
```diff
@@ -766,6 +776,48 @@ class MEOptions:
...
+    """
+    Function to authenticate a user to validate login credentials before
+    running any other API calls. Since we don't have API_PORT setup yet
+    we need to make a test. Strip down output so it doesn't print a
+    big exception and scare users. This function is going to be used
+    before we have the certificates from katello-consumer rpm installed,
+    so we need to disable SSL verification since it's now enabled
+    per default in RHEL7.4+.
+    """
+    port = guess_api_port(options.foreman_fqdn)
```
evgeni: instead of doing this here as a special case, why not using `API_PORT = guess_api_port()` in https://github.com/bergsjoh/katello-client-bootstrap/blob/d1a85da6d7cc9fc4c78bf56c1e325912a7cbca46/bootstrap.py#L851?

and then that should probably be moved *after* the argument parsing, as otherwise `foreman_fqdn` won't be defined yet.

bootstrap.py
```diff
@@ -766,6 +776,48 @@ class MEOptions:
...
+
+    myurl = "https://" + options.foreman_fqdn + ":" + port + "/api/v2/organizations/"
+    try:
+        call_api(myurl, no_verify_ssl=True, silent=True)
+    except:
+        print "ERROR, could not authenticate username. Please try again."
+        sys.exit(2)
+
+
+def guess_api_port(host):
```
evgeni: the function returns an int, this will break the string concatenation above.

evgeni (changes requested):

evgeni:
This looks good to me now. I did not run the code yet, though.

Could you please rebase on the latest master changes?

bootstrap.py
```diff
@@ -1013,6 +1061,8 @@ def exec_service(service, command, failonerror=True):
     # > Clean the environment from LD_... variables
     clean_environment()
 
+    validate_login()
```
evgeni, line 1064: Can you add a comment like `# > Validate credentials`, this will ease the generation of the flow as described in https://github.com/Katello/katello-client-bootstrap/blob/master/CONTRIBUTING.md#developer-and-contributor-notes

bergsjoh:
Rebased.

evgeni (changes requested):
one small change request, but otherwise looks fine.

evgeni:
after discussion with @bergsjoh on IRC: this does not work as intended, as Katello/Satellite also has 8443 open (Candlepin). Going to have to hit `/katello/api/ping` or similar.

bergsjoh:
Bleh, guess_api_port() won't work because satellite is also listening on 8443 tomcat is running there.. 

What a mess. Will have to solve it some other way.

beav:
@bergsjoh @evgeni  is this still being worked on? It hasn't been updated in awhile.

bergsjoh:
Whoops, almost a year. Sorry I got busy about the time we realized that the first approach didn't work. Haven't looked into the alternate of using ping or so still. If you want to finish it up, feel free @beav.

sideangleside:
@beav @bergsjoh I'll rebase, improve and finish this PR unless anyone has any aversions.

bergsjoh:
Please go ahead @sideangleside. Thx.

evgeni:
@sideangleside when I last talked with @bergsjoh about this, we ended up in a dark corner: the credential verification has to happen *before* we install the `katello-ca-consumer` RPM, so we don't know the port yet. both Katello and Content Proxy have 443 *and* 8443 open, but only one of them is serving the right API.

So you'd need to do a `guess_port()`, which connects to both ports and tries an (unathenticated) ping, and then use the port that answered correctly in the rest of the code.

Good luck :)

bootstrap.py
```diff
@@ -804,6 +836,20 @@ def exec_service(service, command, failonerror=True):
             exec_failok("/sbin/service %s %s" % (service, command))
 
 
+def guess_api_port(host):
```
evgeni, line 839: I think I'd go with the following version:

```python
def guess_api_port(hostname):
    """Helper function to get the API port by probing"""
    for port in [443, 8443]:
        url = 'https://' + hostname + ':' + str(port) + '/katello/api/status'
        try:
            call_api(url, no_verify_ssl=True, silent=True, safe=True)
            return str(port)
        except:  # noqa: E722, pylint:disable=bare-except
            pass
    return "443"
```

the `/katello/api/status` endpoint doesn't require authentication and returns a JSON.

We just need to make `call_api` not to call `sys.exit(1)` if `safe=True`

Issue #73, opened 2016-05-13
release is not used


evgeni:
`register_systems(org_name, activationkey, release)` gets a `release` parameter but does not use it (that was removed in b63a0cea7f25b29d7a25f610d680da61b80e0f41).

also, while at it, the RHEL5 check probably should use `options.release` in https://github.com/Katello/katello-client-bootstrap/blob/master/bootstrap.py#L495

peppos:
@evgeni I confirm you that "option.release" correctly work  in RHEL5

soukron:
https://github.com/Katello/katello-client-bootstrap/pull/205 to fix this. I faced the problem today and it's straight forward to add it.

This was my scenario to reproduce it:
 1 create a CV with RHEL 7.3 repository (not RHEL 7Server)
 2 boot a RHEL 7.3 VM and subscribe it manually and run yum clean && yum repolist
    2.1 repos URL use /7Server/ in their release part which is not in the CV
 3 use subscription-manager release --set 7.3 and run yum clean && yum repolist
    3.1 repos URL use /7.3/ in their release part, which is correct
 4 use bootstrap.py to register the server again without --release
    4.1 katello-agent installation fails because 2.1
 5 use bootstrap.py to register the server again but use --release 7.3
    5.1 katello-agent installation fails because --release is not set at registration time, so 2.1 again

soukron:
Also note that I know the fact that the activation kay may also set the release, but given the existence of the parameter I'd prefer to be able to use it to force different releases and not being forced to create multiple activation keys.

sideangleside:
Should we be even allowing users to override the release version of their activation keys?. When using multiple activation keys, you have to contend with the behavior of multiple activation keys (the key that is the furthest to the right (on the CLI) wins), and now you are giving the user the ability to override that. That implies that the admin bootstrapping the client is smarter than the activation key, which seems incorrect. 

Most of the `bootstrap.py` options either 

- Provide data that is required that isn't provided by the server (such as the server URL, Org, Location, Operating system)  **OR**
- Change the behavior of the script itself. (like the --skip options, --install-packages, etc)

This plus the fact that `rhn-migrate-classic-to-rhsm` doesnt support `--release`, I am in the camp that we shouldn't implement it.

soukron:
Yeah, that's another option: get rid of the parameter, but then you're forcing the users to create multiple activation keys instead of just allowing to override by using a method provided by subscription-manager itself. 

[...]

PR #211, opened 2017-08-29
use release when passed via cli

evgeni:
Fixes: #73


bootstrap.py
```diff
@@ -977,6 +981,7 @@ def exec_service(service, command, failonerror=True):
         print "PARTITIONTABLE - %s" % options.partitiontable
         print "ORG - %s" % options.org
         print "ACTIVATIONKEY - %s" % options.activationkey
+        print "CONTENT RELEASE" % options.release
```
sideangleside: should be ` print "CONTENT RELEASE - %s" % options.release`

sideangleside (changes requested):
Other than this one minor change, this LGTM.

bootstrap.py
evgeni: ugh, right :face_with_head_bandage:

evgeni:
fixed.

PR #218, opened 2017-09-07
attempt graceful unregister when migrating from RHSM/SAM. Fixes #217

### LINKED ISSUE #217: [RFE] Migrating a system from RHSM or SAM using the bootstrap script should not require the --force option.
Description of problem from [bz 1478769](https://bugzilla.redhat.com/show_bug.cgi?id=1478769)

When migrating a system from RHSM or SAM using bootstrap.py, the script will fail as it believes the system is already registered and the underlying subscription-manager command fails. Passing the --force option to bootstrap.py allows the registration to complete, however, this behavior is inconsistent. migrating from RHN/Sat doesn't require `--force`; migrating from RHSM/SAM does. 

Additionally, the --force option was originally intended to allow a user to override the logic that bootstrap.py has to prevent a user from running the script against a system which is already properly registered. Having to use the --force switch for a common registration workflow is counter-intuitive and overloads the meaning of --force

the script fails with an error similar to the following
~~~
[NOTIFICATION], [2017-08-03 06:51:26], [Retrieving Client CA Certificate RPMs] 
[RUNNING], [2017-08-03 06:51:26], [rpm -Uvh http://satellite.example.com/pub/katello-ca-consumer-latest.noarch.rpm] 
Retrieving http://satellite.example.com/pub/katello-ca-consumer-latest.noarch.rpm
Preparing...                ##################################################
katello-ca-consumer-satellit##################################################
[SUCCESS], [2017-08-03 06:51:27], [rpm -Uvh http://satellite.example.com/pub/katello-ca-consumer-latest.noarch.rpm], completed successfully.

[NOTIFICATION], [2017-08-03 06:51:27], [Writing FQDN katello-fact] 
[NOTIFICATION], [2017-08-03 06:51:27], [Calling subscription-manager] 
[RUNNING], [2017-08-03 06:51:27], [/usr/sbin/subscription-manager register --org 'Example' --name 'sam-client.example.com' --activationkey 'ak_Reg_To_Library_EL6'  --serverurl=https://satellite.example.com:443/rhsm --baseurl=https://satellite.example.com/pulp/repos] 
[ERROR], [2017-08-03 06:51:28], EXITING: [/usr/sbin/subscription-manager register --org 'Example' --name 'sam-client.example.com' --activationkey 'ak_Reg_To_Library_EL6'  --serverurl=https://satellite.example.com:443/rhsm --baseurl=https://satellite.example.com/pulp/repos] failed to execute properly.
This system is already registered. Use --force to override
~~~

Expected results:
Register should complete without having to pass --force.

sideangleside:
Fixes #217

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -230,7 +230,9 @@ def get_bootstrap_rpm():
         print_generic("A Katello CA certificate is already installed. Assuming system is registered")
         print_generic("To override this behavior, run the script with the --force option. Exiting.")
         sys.exit(1)
-
+    if os.path.exists('/etc/pki/consumer/cert.pem'):
+        print_generic('System appears to be registered via another entitlement server. Attempting unregister')
+        unregister_system()
     if options.download_method == "https":
         print_generic("Writing custom cURL configuration to allow download via HTTPS without certificate verification")
         curl_config_dir = tempfile.mkdtemp()
@@ -318,8 +320,6 @@ def register_systems(org_name, activationkey):
     options.smargs += " --serverurl=https://%s:%s/rhsm --baseurl=https://%s/pulp/repos" % (options.foreman_fqdn, API_PORT, options.foreman_fqdn)
     if options.force:
         options.smargs += " --force"
-        exec_failok("subscription-manager unregister")
-        exec_failok("subscription-manager clean")
     if options.release:
         options.smargs += " --release %s" % options.release
     exec_failexit("/usr/sbin/subscription-manager register --org '%s' --name '%s' --activationkey '%s' %s" % (org_label, FQDN, activationkey, options.smargs))
@@ -329,7 +329,8 @@ def register_systems(org_name, activationkey):
 def unregister_system():
     """Unregister the host using `subscription-manager`."""
     print_generic("Unregistering")
-    exec_failexit("/usr/sbin/subscription-manager unregister")
+    exec_failok("/usr/sbin/subscription-manager unregister")
+    exec_failok("/usr/sbin/subscription-manager clean")
 
 
 def clean_katello_agent():


evgeni:
I am trying to understand the various situations that can happen:
1. System is vanilla, just trying to freshly register
1. System was on Spacewalk/Satellite5/RHN and is trying to migrate
1. System was on Katello/Satellite6 and is trying to migrate
1. System was on RHSM and is trying to migrate

So:
1. does not need anything, and especially no `--force`. Neither of the files we're checking are there and thus `bootstrap.py` will just register fine.
2. is handled using `migrate_systems()` and `rhn-migrate-classic-to-rhsm`. Neither of the "new" files we're checking are there and we're fine again.
3. Both `/etc/rhsm/ca/katello-server-ca.pem` and `/etc/pki/consumer/cert.pem` exist. We error out if no `--force` was given. If `--force` was given, we yank all the Katello-related files and try to re-register. With your change we would also call "unregister", but as the CA certs are gone, that will fail (not a problem).
4. Only `/etc/pki/consumer/cert.pem` exists. Previously, we would hit an error if no `--force` was given. Now you detect that case and try to unregister and clean.

When the above is correct, then I think:
* the unregister/clean call in `register_system` is now obsolete (https://github.com/sideangleside/katello-client-bootstrap/blob/b9ea1011cc16c6f2ba376768e5e4baa336239feb/bootstrap.py#L325-L326) as you have called unregister/clean earlier already.
* We have `unregister_system()` which we only use when doing `remove`. Shall we extend it to call both unregister and clean and use it instead of duplicating code?

sideangleside:
>  the unregister/clean call in register_system is now obsolete (https://github.com/sideangleside/katello-client-bootstrap/blob/b9ea1011cc16c6f2ba376768e5e4baa336239feb/bootstrap.py#L325-L326) as you have called unregister/clean earlier already.

+1 

>  We have unregister_system() which we only use when doing remove. Shall we extend it to call both unregister and clean and use it instead of duplicating code?

I guess this depends. `unregister_system()` calls `subscription_manager` via exec_failexit, which I believe is the correct behavior. (Fail if we can't remove). Though I guess with the added `subscription-manager clean` the differences are academic at this point, because `subscription-manager clean` will always (short of a _very_ dire problem with the system), be successful.

Rocco83:
> I am trying to understand the various situations that can happen:
> 
>     System is vanilla, just trying to freshly register
>     System was on Spacewalk/Satellite5/RHN and is trying to migrate
>     System was on Katello/Satellite6 and is trying to migrate
>     System was on RHSM and is trying to migrate

I would add:
System is on Katello/Satellite6 and i want to fix the installation.
But again, not in this PR.
See https://github.com/Katello/katello-client-bootstrap/issues/221

evgeni (approved):
ACK

Issue #210, opened 2017-08-29
Define AuthorizedKeysFile path when rex is enable


alosadagrande:
There are some installations where AuthorizedKeysFile value is not the default one ($HOME/.ssh/authorized_keys). In thoses cases rex goes OK but obviously as the sshd is expecting the ssh pub keys in other directory, ssh with pubkey auth is failing.

alosadagrande:
I have been taking a look and seems that there is no environment variable which can help us to locate the proper path for user's AuthorizedKeysFile.

I see two options:

1. Use regular expressions to extract the value from /etc/ssh/sshd_config file. Probably a if/else structure must be coded as well due to different values like: none, .ssh/authorized_keys, /etc/keys/%u/authorized_keys or even a single file 
2. Allow user to set this value during execution as --rex-authkey-path. Default value should be $HOME/.ssh/authorized_keys as it is already coded.

evgeni:
I'd vote for 2. I hate guessing ;)

sideangleside:
Same. +1 to option 2. (And I am afraid to ask who doesn't use the standard path for the AuthorizeKeysFile) :)

evgeni:
I do, on boxes where I don't want the users to be able to change their keys.

alosadagrande:
First time I see... but seems pretty reasonable from a security point of view.

Issue #204, opened 2017-08-21
Define multiple capsules to copy ssh keys when enabling rex


alosadagrande:
When enabling remote execution (--rex) registered client downloads the foreman-proxy ssh key from the --server value defined. In case you have several capsules on the same location it is useful to allow to pass this information.

Could be useful to create a new parameter called --rex-capsules which overrides the --server value when added to boostrap.py. Also it should allow to enter comma separated values in case of multiple capsules at the same location.

sideangleside:
Assuming we implement as designed, what is the failure mode: should we fail

- if we cannot download _any_ of the keys
- if we cannot download _all_ of the keys

Would a better solution be to allow the user to just provide an arbitrary URL which contains the public keys concatenated?

alosadagrande:
Actually I already coded as I suggested at the beginning. Basically using the same function "install_foreman_ssh_key" looping each of the capsules passed by parameter (--rex-capsules). So taking a look at the function it just prints a warning / generic exception if cannot connect or if there is any problem. 

**/tools/scripts/satellite/bootstrap-wrapper.sh: [NOTIFICATION], [2017-08-29 15:56:01], [Could not reach the server. Error: [Errno 110] Connection timed out]** 
/tools/scripts/satellite/bootstrap-wrapper.sh: [NOTIFICATION], [2017-08-29 15:56:01], [Foreman's SSH key/s was/were added to /root/.ssh/authorized_keys] 
/tools/scripts/satellite/bootstrap-wrapper.sh: [NOTIFICATION], [2017-08-29 15:56:01], [Foreman's SSH key/s was/were added to /root/.ssh/authorized_keys] 



However as you suggested it makes now more sense to me to just create a concatenated file of all foreman's key ready to be imported. The aim of this issue is to solve the problem when more than one capsule is at the same location, probably they're behind a load balancer. In that case it will make sense to previously concatenate the keys and just call the load balancer VIP which will be in charge of retrieve the file from whatever capsule is online. If there is an offline capsule, we have to assume the LB is smart enough to route the call to the online one - otherwise would be a networking configuration problem ;) 

In my opinion the drawback is that you have to create the concatenated key yourself - out of the bootstrap.py. It's an easy task to automate and it's a one-time-shot, but must be taken into account.

I would suggest option called --rex-url-keys which would accept an http/https URI.

PR #220, opened 2017-09-08
Add multiple REX options. See issues #204 #210

alosadagrande:
Add multiple options to remote execution in order to solve enhancement #204 and #210:

--rex-capsules: Comma separated list of capsules to install Foreman's SSH keys for remote execution. It will iterate all over the list of capsules. If one fails user is notified by the task keeps going until all capsules were tried.

--rex-urlkeyfile: HTTP/S location to install a file containing one or multiple Foreman's SSH keys for remote execution. Instead of iterate you can create file on /pub folder  that contains all ssh public keys of several capsules and tell bootstrap.py to fech it and install on the local user authorized keys file.

--rex-authpath. Local folder where is placed the authorized_keys file to install Foreman's SSH keys for remote execution. Default ~/.ssh. I found that several customers specifies AuthorizedKeysFile on sshd_config on a different path.

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -514,19 +514,24 @@ def fully_update_the_box():
 
 # curl https://satellite.example.com:9090/ssh/pubkey >> ~/.ssh/authorized_keys
 # sort -u ~/.ssh/authorized_keys
-def install_foreman_ssh_key():
+def install_foreman_ssh_key(remote_url):
     """
     Download and install the Satellite's SSH public key into the foreman user's
-    authorized keys file, so that remote execution becomes possible.
+    authorized keys file location, so that remote execution becomes possible.
+    If not set default is ~/.ssh/authorized_keys
     """
-    userpw = pwd.getpwnam(options.remote_exec_user)
-    foreman_ssh_dir = os.sep.join([userpw.pw_dir, '.ssh'])
-    foreman_ssh_authfile = os.sep.join([foreman_ssh_dir, 'authorized_keys'])
-    if not os.path.isdir(foreman_ssh_dir):
-        os.mkdir(foreman_ssh_dir, 0700)
-        os.chown(foreman_ssh_dir, userpw.pw_uid, userpw.pw_gid)
+    if not options.remote_exec_authpath:
+        userpw = pwd.getpwnam(options.remote_exec_user)
+        options.remote_exec_authpath = os.path.join(userpw.pw_dir, '.ssh', 'authorized_keys')
+        foreman_ssh_dir = os.path.join(userpw.pw_dir, '.ssh')
+        if not os.path.isdir(foreman_ssh_dir):
+            os.mkdir(foreman_ssh_dir, 0700)
+            os.chown(foreman_ssh_dir, userpw.pw_uid, userpw.pw_gid)
+    elif not os.path.isfile(options.remote_exec_authpath):
+        print_error("Foreman's SSH key not installed. File where authorized_keys must be located is not found: %s" % options.remote_exec_authpath)
+        return
     try:
-        foreman_ssh_key = urllib2.urlopen(("https://%s:9090/ssh/pubkey" % options.foreman_fqdn), timeout=options.timeout).read()
+        foreman_ssh_key = urllib2.urlopen(remote_url, timeout=options.timeout).read()
     except urllib2.HTTPError, e:
         print_generic("The server was unable to fulfill the request. Error: %s - %s" % (e.code, e.reason))
         print_generic("Please ensure the Remote Execution feature is configured properly")
@@ -535,14 +540,14 @@ def install_foreman_ssh_key():
     except urllib2.URLError, e:
         print_generic("Could not reach the server. Error: %s" % e.reason)
         return
-    if os.path.isfile(foreman_ssh_authfile):
-        if foreman_ssh_key in open(foreman_ssh_authfile, 'r').read():
-            print_generic("Foreman's SSH key is already present in %s" % foreman_ssh_authfile)
+    if os.path.isfile(options.remote_exec_authpath):
+        if foreman_ssh_key in open(options.remote_exec_authpath, 'r').read():
+            print_generic("Foreman's SSH key already present in %s" % options.remote_exec_authpath)
             return
-    output = os.fdopen(os.open(foreman_ssh_authfile, os.O_WRONLY | os.O_CREAT, 0600), 'a')
+    output = os.fdopen(os.open(options.remote_exec_authpath, os.O_WRONLY | os.O_CREAT, 0600), 'a')
     output.write(foreman_ssh_key)
-    os.chown(foreman_ssh_authfile, userpw.pw_uid, userpw.pw_gid)
-    print_generic("Foreman's SSH key was added to %s" % foreman_ssh_authfile)
+    os.chown(options.remote_exec_authpath, userpw.pw_uid, userpw.pw_gid)
+    print_generic("Foreman's SSH key added to %s" % options.remote_exec_authpath)
     output.close()
 
 
@@ -997,6 +1002,9 @@ if __name__ == '__main__':
     parser.add_option("--unmanaged", dest="unmanaged", action="store_true", help="Add the server as unmanaged. Useful to skip provisioning dependencies.")
     parser.add_option("--rex", dest="remote_exec", action="store_true", help="Install Foreman's SSH key for remote execution.", default=False)
     parser.add_option("--rex-user", dest="remote_exec_user", default="root", help="Local user used by Foreman's remote execution feature.")
+    parser.add_option("--rex-proxies", dest="remote_exec_proxies", help="Comma separated list of proxies to install Foreman's SSH keys for remote execution.")
+    parser.add_option("--rex-urlkeyfile", dest="remote_exec_url", help="HTTP/S location to install a file containing one or multiple Foreman's SSH keys for remote execution.")
+    parser.add_option("--rex-authpath", dest="remote_exec_authpath", help="Full path to local authorized_keys file in order to install Foreman's SSH keys for remote execution. Default ~/.ssh/authorized_keys")
     parser.add_option("--enablerepos", dest="enablerepos", help="Repositories to be enabled via subscription-manager - comma separated", metavar="enablerepos")
     parser.add_option("--skip", dest="skip", action="append", help="Skip the listed steps (choices: %s)" % SKIP_STEPS, choices=SKIP_STEPS, default=[])
     parser.add_option("--ip", dest="ip", help="IPv4 address of the primary interface in Foreman (defaults to the address used to make request to Foreman)")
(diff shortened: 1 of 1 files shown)


bootstrap.py
```diff
@@ -880,6 +889,9 @@ def exec_service(service, command, failonerror=True):
     parser.add_option("--unmanaged", dest="unmanaged", action="store_true", help="Add the server as unmanaged. Useful to skip provisioning dependencies.")
     parser.add_option("--rex", dest="remote_exec", action="store_true", help="Install Foreman's SSH key for remote execution.", default=False)
     parser.add_option("--rex-user", dest="remote_exec_user", default="root", help="Local user used by Foreman's remote execution feature.")
+    parser.add_option("--rex-capsules", dest="remote_exec_capsules", default="", help="Comma separated list of capsules to install Foreman's SSH keys for remote execution.")
```
evgeni: If there is no default, you can omit the `default` key completely.

bootstrap.py
```diff
@@ -449,12 +458,12 @@ def install_foreman_ssh_key():
         return
     if os.path.isfile(foreman_ssh_authfile):
         if foreman_ssh_key in open(foreman_ssh_authfile, 'r').read():
-            print_generic("Foreman's SSH key is already present in %s" % foreman_ssh_authfile)
+            print_generic("Foreman's SSH key/s already present in %s" % foreman_ssh_authfile)
```
evgeni: why this change? do you want the sentence to be "ssh key (or keys) already present"? given the function still adds one key at a time, I think we should not change that. But we could add the remote-url to indicate which key was present?

bootstrap.py
```diff
@@ -449,12 +458,12 @@ def install_foreman_ssh_key():
...
             return
     output = os.fdopen(os.open(foreman_ssh_authfile, os.O_WRONLY | os.O_CREAT, 0600), 'a')
     output.write(foreman_ssh_key)
     os.chown(foreman_ssh_authfile, userpw.pw_uid, userpw.pw_gid)
-    print_generic("Foreman's SSH key was added to %s" % foreman_ssh_authfile)
+    print_generic("Foreman's SSH key/s added to %s" % foreman_ssh_authfile)
```
evgeni: same as above with s/present/added/ ;)

bootstrap.py
```diff
@@ -426,19 +426,28 @@ def fully_update_the_box():
...
+            return
+    else:
+        foreman_ssh_dir = os.sep.join([userpw.pw_dir, '.ssh'])
+        foreman_ssh_authfile = os.sep.join([foreman_ssh_dir, 'authorized_keys'])
     if not os.path.isdir(foreman_ssh_dir):
         os.mkdir(foreman_ssh_dir, 0700)
         os.chown(foreman_ssh_dir, userpw.pw_uid, userpw.pw_gid)
     try:
-        foreman_ssh_key = urllib2.urlopen("https://%s:9090/ssh/pubkey" % options.foreman_fqdn).read()
+        foreman_ssh_key = urllib2.urlopen("%s" % remote_url).read()
```
evgeni: `foreman_ssh_key = urllib2.urlopen(remote_url).read()`

no need for string format

bootstrap.py
```diff
@@ -426,19 +426,28 @@ def fully_update_the_box():
...
-    foreman_ssh_authfile = os.sep.join([foreman_ssh_dir, 'authorized_keys'])
+    if options.remote_exec_authpath:
+        if os.path.isdir(options.remote_exec_authpath):
+            foreman_ssh_dir = options.remote_exec_authpath
+            foreman_ssh_authfile = options.remote_exec_authpath + "/authorized_keys"
+        else:
+            print_error("Foreman's SSH keys not installed. Directory where authorized_keys file must be located is not found: %s" % options.remote_exec_authpath)
+            return
+    else:
+        foreman_ssh_dir = os.sep.join([userpw.pw_dir, '.ssh'])
```
evgeni: can we make the default value of `options.remote_exec_authpath` to be `os.path.join(userpw.pw_dir, '.ssh')` then we could skip this if/else block.

note that I changed `os.sep.join([userpw.pw_dir, '.ssh'])` to `os.path.join(userpw.pw_dir, '.ssh')` which I think is more correct here.

bootstrap.py
```diff
@@ -426,19 +426,28 @@ def fully_update_the_box():
...
+    authorized keys file location, so that remote execution becomes possible.
+    By default is ~/.ssh/authorized_keys
     """
     userpw = pwd.getpwnam(options.remote_exec_user)
-    foreman_ssh_dir = os.sep.join([userpw.pw_dir, '.ssh'])
-    foreman_ssh_authfile = os.sep.join([foreman_ssh_dir, 'authorized_keys'])
+    if options.remote_exec_authpath:
+        if os.path.isdir(options.remote_exec_authpath):
+            foreman_ssh_dir = options.remote_exec_authpath
+            foreman_ssh_authfile = options.remote_exec_authpath + "/authorized_keys"
```
evgeni: There is no requirement for the file to be called `authorized_keys`. In fact, some of my boxes have `AuthorizedKeysFile /etc/ssh/authorized_keys/%u` (with `%u` being the username). So I think we just should take the full path at the command line and not alter it at all.

evgeni (changes requested):

bootstrap.py
alosadagrande: OK

bootstrap.py
alosadagrande: Actually from the first comment options.remote_exec_authpath is the complete path to authorizedKey file not a dir.

bootstrap.py
alosadagrande: ok

bootstrap.py
alosadagrande: OK. Depending on the rex option you're adding one key at a time or multiple when copying a remote file containing multiples keys. So if adding multiples you're checking multiple keys at a time.

bootstrap.py
alosadagrande: OK

bootstrap.py
alosadagrande: ok. same

bootstrap.py
```diff
@@ -426,19 +426,24 @@ def fully_update_the_box():
...
-        os.chown(foreman_ssh_dir, userpw.pw_uid, userpw.pw_gid)
+    if not options.remote_exec_authpath:
+        userpw = pwd.getpwnam(options.remote_exec_user)
+        options.remote_exec_authpath = os.path.join(*[userpw.pw_dir, '.ssh', 'authorized_keys'])
+        foreman_ssh_dir = os.path.join(*[userpw.pw_dir, '.ssh'])
+        if not os.path.isdir(foreman_ssh_dir):
+            os.mkdir(foreman_ssh_dir, 0700)
+            os.chown(foreman_ssh_dir, userpw.pw_uid, userpw.pw_gid)
+    elif not os.path.isfile(options.remote_exec_authpath):
+        print_error("Foreman's SSH key not installed. File where authorized_keys must be located is not found: %s" % options.remote_exec_authpath)
```
evgeni, line 531: Why? We can just create the file (and we actually do)?

bootstrap.py
```diff
@@ -880,6 +885,9 @@ def exec_service(service, command, failonerror=True):
     parser.add_option("--unmanaged", dest="unmanaged", action="store_true", help="Add the server as unmanaged. Useful to skip provisioning dependencies.")
     parser.add_option("--rex", dest="remote_exec", action="store_true", help="Install Foreman's SSH key for remote execution.", default=False)
     parser.add_option("--rex-user", dest="remote_exec_user", default="root", help="Local user used by Foreman's remote execution feature.")
+    parser.add_option("--rex-capsules", dest="remote_exec_capsules", help="Comma separated list of capsules to install Foreman's SSH keys for remote execution.")
+    parser.add_option("--rex-urlkeyfile", dest="remote_exec_url", help="HTTP/S location to install a file containing one or multiple Foreman's SSH keys for remote execution.")
+    parser.add_option("--rex-authpath", dest="remote_exec_authpath", help="Local folder where is placed the authorized_keys file to install Foreman's SSH keys for remote execution. Default ~/.ssh")
```
evgeni: Local file, not folder :)

bootstrap.py
```diff
@@ -426,19 +426,24 @@ def fully_update_the_box():
...
     """
-    userpw = pwd.getpwnam(options.remote_exec_user)
-    foreman_ssh_dir = os.sep.join([userpw.pw_dir, '.ssh'])
-    foreman_ssh_authfile = os.sep.join([foreman_ssh_dir, 'authorized_keys'])
-    if not os.path.isdir(foreman_ssh_dir):
-        os.mkdir(foreman_ssh_dir, 0700)
-        os.chown(foreman_ssh_dir, userpw.pw_uid, userpw.pw_gid)
+    if not options.remote_exec_authpath:
+        userpw = pwd.getpwnam(options.remote_exec_user)
+        options.remote_exec_authpath = os.path.join(*[userpw.pw_dir, '.ssh', 'authorized_keys'])
```
evgeni: `os.path.join(userpw.pw_dir, '.ssh', 'authorized_keys')` should do it too, no?

bootstrap.py
```diff
@@ -426,19 +426,24 @@ def fully_update_the_box():
...
+        foreman_ssh_dir = os.path.join(*[userpw.pw_dir, '.ssh'])
```
evgeni: same as above

evgeni (dismissed):

bootstrap.py
alosadagrande: That's the case when the path is entered by user. I think must be warned.

If there is no remote_exec_autpath we pick the default one (~/.ssh/authorized_keys) where we create the folder if needed

alosadagrande: We should not create the complete path to the remote_exec_autpath if it's entered by user, otherwise we'll probably face security issues in the future.

evgeni:
rebased and squashed this (thanks @fcami for the rebase work)

evgeni:
@sideangleside as I got my hands dirty in here, you'll have to do the final review

beav:
@alosadagrande it looks like this needs a rebase

fcami:
@beav @alosadagrande @evgeni I'll rebase

fcami:
@beav @evgeni could you please review and pull from https://github.com/fcami/katello-client-bootstrap/tree/multi-key-rex - it seems OK to me but I haven't tested it yet.

beav:
@fcami the change looks OK to me but I didn't test it either.

If @alosadagrande doesn't update this PR, the best bet is for you to create a new branch, then do "git cherry-pick -x 299b0349c3168e0cf0a5473e2a1a3e967b98a0ef" and resolve the conflict, then do "git add bootstrap.py" and "git commit". That way it will keep the authorship information correct and you can then create a new PR and we can close this one out.

fcami:
@beav thanks - please see https://github.com/Katello/katello-client-bootstrap/pull/253 then

evgeni:
obsoleted by https://github.com/Katello/katello-client-bootstrap/pull/281

PR #227, opened 2017-11-22
Capsule migration option, take 2

wzzrd:
So after a long pause, here's attempt two to bring in a capsule migration option into bootstrap.py.

Because the change is fairly large, here's the breakdown, to start with the nitty-gritty stuff:
- changes were required to basic functions, like `put_json` and `get_bootstrap_rpm`. The `put_json` function needed to accept a payload to update a host record, and the `get_bootstrap_rpm` needed to have options to override both whether or not to remove existing katello-ca-consumer RPMs, and whether or not to unregister when forcing the installation of new katello-ca-consumer RPMs. Unregistering, needless to say, from a migration PoV, would be bad.
- I also have added an `update_host_capsule_mapping` function to DRY when updating the proxy_ids for a system.
- Some obvious changes have been made to the parts of the script that provide help and figure out what to do with all the options.
- The part that does the actual migration is fairly long, that can't be helped much, I'm afraid. What happens is as follows:
  - get the new katello-ca-consumer from the target capsule and figure out the API_PORT, capsule_id, host_id and the features the capsule offers
  - optionally (depending on whether the `--hostgroup` and `--location` parameters are passed) switch to new location, hostgroup
  - optionally (depending on `puppet` and `foreman` in options.skip) change puppet_proxy_id and puppet_ca_proxy_id
  - optionally (depending on `foreman` in options.skip and the features the new capsule offers) change openscap_proxy_id
  - switch content_source_id
  - re-enable rhsmcertd
  - optionally, depending on options.skip, reconfigure Puppet

Remarks:
- Right now, I check for `foreman` and `puppet` in options.skip when changing puppet_proxy_id, puppet_ca_proxy_id and re-configuring Puppet. I don't think it makes sense to reconfigure Puppet if we skip foreman, because submitted reports will create / update an existing foreman host anyway. So if we skip `puppet` or `foreman`, we don't touch anything Puppet related. Agreed?
- We cannot check whether or not a host has an openscap_proxy_id configured through the API. Therefore, what I do is:
  - if we do not skip foreman and the capsule offers OpenSCAP, we set the openscap_proxy_id. After all, if we are staying in the same hostgroup and that hostgroup has OpenSCAP, we want to switch capsules. If we switch hostgroups to a hostgroup that has OpenSCAP, we want to switch capsules. If our hostgroup does not have OpenSCAP, setting the openscap_proxy_id doesn't matter.
  - if there is no openscap feature on the new capsule, we don't touch openscap_proxy_id, but warn the user. Same reasons as above. 

Sample usage is provided in README.md. Tested on RHEL6 and RHEL7 w/ Satellite 6.2.12. RHEL5 untested, but I'm not using any functionality in Python that wasn't available on Python 2.4 (I think).

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,9 @@ bootstrap Script for migrating existing running systems to Foreman with the Kate
 This script can take a system that is registered to Spacewalk, Satellite 5, Red Hat
 Network Classic and get it registered to Foreman & Katello.
 
+Optionally, you can also move systems between Capsules (both internal and
+external) of one Katello installation by using the `--new-capsule` option.
+
 # What does the Script do?
 
 * Identify which systems management platform is the system registered to (Classic/Sat5 or None)  then perform the following
@@ -20,6 +23,18 @@ Network Classic and get it registered to Foreman & Katello.
 * registering the system to Foreman
 * Configuring the system with a proper Puppet configuration pointing at Foreman
 * Removing/disabling old RHN Classic packages/daemons (rhnsd, osad, etc)
+
+## System already registered to a Foreman + Katello server / Capsule (--new-capsule)
+* Clean the existing Katello agent installation
+* Install the Katello consumer RPM for the target Foreman + Katello server / Capsule
+* Install the Katello agent software again, using the configuration for the
+  target Foreman + Katello / Capsule server
+* Make API calls to switch the system to a different hostgroup (optional)
+* Make API calls to update the Puppet master, Puppet CA, content source and
+  OpenSCAP proxy IDs (optional, except for content source)
+* Re-enable rhsmcertd
+* Update the Puppet configuration for the system to point to the right capsule (optional)
+* Restart Puppet and call for the user to go sign the CSR
 
 ## System not registered to any Red Hat Systems Management Platform:
 
@@ -168,6 +183,25 @@ By default, bootstrap.py does not delete the system's profile from the legacy pl
 ### Migrating a system from one Foreman + Katello installation to another.
 
 There are times where it is necessary to migrate clients from one Foreman + Katello installation to another. For instance, in lieu of upgrading an older Foreman + Katello installation, you choose to build a new installation in parallel. bootstrap.py can then be used to migrate clients from one Foreman + Katello installation to another. Simply provide the `--force` option, and bootstrap.py will remove the previous `katello-ca-consumer-*` package (from the old system), and will install the `katello-ca-consumer-*` package (from the new system), and continue registration as usual.
+
+### Migrating a system from one Foreman + Katello installation 6 or Capsule / to another in the same infrastructure
+
+In order to manually balance the load over multiple Capsule servers, you might
+want to move some existing systems to newly deployed Capsules. You can easily
+do this by running the bootstrap.py script like the examples below. Mind that
+you still have to manually revoke any Puppet certificates on the old capsules!
+
+~~~
+# ./bootstrap.py -l admin --new-capsule --server capsule.example.com
+~~~
+
+If you want to change the hostgroup and location of the system at the same
+time, run:
+
+~~~
+# ./bootstrap.py -l admin --new-capsule --server capsule.example.com \
+    --hostgroup mygroup --location mylocation
+~~~
 
 ### Enabling additional repositories at registration time.
 
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -256,19 +256,22 @@ def is_fips():
     return fips_status == "1"
 
 
-def get_bootstrap_rpm():
+def get_bootstrap_rpm(clean=False, unreg=True):
     """
     Retrieve Client CA Certificate RPMs from the Satellite 6 server.
     Uses --insecure options to curl(1) if instructed to download via HTTPS
-    If called with --force, calls clean_katello_agent().
+    This function is usually called with clean=options.force, which ensures
+    clean_katello_agent() is called if --force is specified. You can optionally
+    pass unreg=False to bypass unregistering a system (e.g. when moving between
+    capsules.
     """
-    if options.force:
+    if clean:
         clean_katello_agent()
     if os.path.exists('/etc/rhsm/ca/katello-server-ca.pem'):
         print_generic("A Katello CA certificate is already installed. Assuming system is registered")
         print_generic("To override this behavior, run the script with the --force option. Exiting.")
         sys.exit(1)
-    if os.path.exists('/etc/pki/consumer/cert.pem'):
+    if os.path.exists('/etc/pki/consumer/cert.pem') and unreg:
         print_generic('System appears to be registered via another entitlement server. Attempting unregister')
         unregister_system()
     if options.download_method == "https":
@@ -306,6 +309,13 @@ def enable_rhsmcertd():
     """
     enable_service("rhsmcertd")
     exec_service("rhsmcertd", "restart")
+
+
+def is_registered():
+    # Check if all required certificates are in place (i.e. a system is
+    # registered to begin with) before we start changing things
+    return (os.path.exists('/etc/rhsm/ca/katello-server-ca.pem') and
+            os.path.exists('/etc/pki/consumer/cert.pem'))
 
 
 def migrate_systems(org_name, activationkey):
@@ -465,6 +475,13 @@ server          = %s
         puppet_conf.write("""noop            = true
 """)
     puppet_conf.close()
+    noop_puppet_signing_run()
+    if 'puppet-enable' not in options.skip:
+        enable_service("puppet")
+        exec_service("puppet", "restart")
+
+
+def noop_puppet_signing_run():
     print_generic("Running Puppet in noop mode to generate SSL certs")
     print_generic("Visit the UI and approve this certificate via Infrastructure->Capsules")
     print_generic("if auto-signing is disabled")
@@ -591,9 +608,30 @@ def delete_json(url):
     return call_api(url, method='DELETE')
 
 
-def put_json(url):
+def put_json(url, jdata=None):
     """Use `call_api` to place a "PUT" REST API call."""
-    return call_api(url, method='PUT')
+    return call_api(url, data=jdata, method='PUT')
+
+
+def update_host_capsule_mapping(attribute, capsule_id, host_id):
+    url = "https://" + options.foreman_fqdn + ":" + str(API_PORT) + "/api/v2/hosts/" + str(host_id)
+    if attribute == 'content_source_id':
+        jdata = json.loads('{"host": {"content_facet_attributes": {"content_source_id": "%s"}, "content_source_id": "%s"}}' % (capsule_id, capsule_id))
+    else:
+        jdata = json.loads('{"host": {"%s": "%s"}}' % (attribute, capsule_id))
+    return put_json(url, jdata)
+
+
+def get_capsule_features(capsule_id):
+    url = "https://" + options.foreman_fqdn + ":" + str(API_PORT) + "/katello/api/capsules/%s" % str(capsule_id)
+    return [f['name'] for f in get_json(url)['features']]
+
+
+def update_host_config(attribute, value, host_id):
+    attribute_id = return_matching_foreman_key(attribute + 's', 'title="%s"' % value, 'id', False)
+    json_key = attribute + "_id"
+    jdata = json.loads('{"host": {"%s": "%s"}}' % (json_key, attribute_id))
+    put_json("https://" + options.foreman_fqdn + ":" + API_PORT + "/api/hosts/%s" % host_id, jdata)
 
 
 def return_matching_foreman_key(api_name, search_key, return_key, null_result_ok=False):
@@ -948,6 +986,7 @@ if __name__ == '__main__':
     parser.add_option("--deps-repository-url", dest="deps_repository_url", help="URL to a repository that contains the subscription-manager RPMs")
     parser.add_option("--deps-repository-gpg-key", dest="deps_repository_gpg_key", help="GPG Key to the repository that contains the subscription-manager RPMs", default="file:///etc/pki/rpm-gpg/RPM-GPG-KEY-redhat-release")
     parser.add_option("--install-packages", dest="install_packages", help="List of packages to be additionally installed - comma separated", metavar="installpackages")
+    parser.add_option("--new-capsule", dest="new_capsule", action="store_true", help="Switch the server to a new capsule for content and Puppet. Pass --server with the Capsule FQDN as well.")
     parser.add_option("-t", "--timeout", dest="timeout", type="int", help="Timeout (in seconds) for API calls and subscription-manager registration. Defaults to %default", metavar="timeout", default=900)
     (options, args) = parser.parse_args()
 
@@ -972,9 +1011,12 @@ if __name__ == '__main__':
     #     if removing from foreman:
     #        foreman_fqdn
     if not ((options.remove and ('foreman' in options.skip or options.foreman_fqdn)) or
-            (options.foreman_fqdn and options.org and options.activationkey and ('foreman' in options.skip or options.hostgroup))):
-        if not options.remove:
+            (options.foreman_fqdn and options.org and options.activationkey and ('foreman' in options.skip or options.hostgroup)) or
+            (options.foreman_fqdn and options.new_capsule)):
+        if not options.remove and not options.new_capsule:
             print "Must specify server, login, organization, hostgroup and activation key.  See usage:"
+        elif options.new_capsule:
+            print "Must use both --new-capsule and --server. See usage:"
         else:
             print "Must specify server.  See usage:"
         parser.print_help()
@@ -1083,8 +1125,8 @@ if __name__ == '__main__':
     # > Clean the environment from LD_... variables
     clean_environment()
 
-    # > IF RHEL 5 and not removing, prepare the migration.
-    if not options.remove and int(RELEASE[0]) == 5:
+    # > IF RHEL 5, not removing, and not moving to new capsule prepare the migration.
+    if not options.remove and int(RELEASE[0]) == 5 and not options.new_capsule:
         prepare_rhel5_migration()
 
     if options.remove:
@@ -1108,7 +1150,7 @@ if __name__ == '__main__':
         print_generic('This system is registered to RHN. Attempting to migrate via rhn-classic-migrate-to-rhsm')
         install_prereqs()
         check_migration_version()
-        get_bootstrap_rpm()
+        get_bootstrap_rpm(clean=options.force)
         generate_katello_facts()
         API_PORT = get_api_port()
         if 'foreman' not in options.skip:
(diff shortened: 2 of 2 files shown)


wzzrd:
Soooo..... the bare excepts that Travis still complains about have been in there for months and have your prints all over them, @evgeni ;)

Not sure what to do about those...

evgeni:
@wzzrd ah, this is new in flake8 3.5.0. Let's see how we can fix that.

wzzrd:
<polite voice>@evgeni, @sideangelside, so, any comments on how to move this forward, i.e. get this merged?</polite voice>

bootstrap.py
```diff
@@ -1053,12 +1066,78 @@ def exec_service(service, command, failonerror=True):
         migrate_systems(options.org, options.activationkey)
         if options.enablerepos:
             enable_repos()
+    elif options.new_capsule and options.foreman_fqdn:
```
evgeni: `foreman_fqdn` should be always set at this point, no?

bootstrap.py
```diff
@@ -1053,12 +1066,78 @@ def exec_service(service, command, failonerror=True):
...
+        # > MANUAL SIGNING OF CSR OR MANUALLY CREATING AUTO-SIGN RULE STILL REQUIRED!
+        # > API doesn't have a public endpoint for creating auto-sign entries yet!
+
+        # Make system ready for switch, gather required data
+        get_bootstrap_rpm(clean=True, unreg=False)
+        install_katello_agent()
+        API_PORT = get_api_port()
+        capsule_id = return_matching_foreman_key('smart_proxies', 'name="%s"' % options.foreman_fqdn, 'id', False)
+        host_id = return_matching_foreman_key('hosts', 'name="%s"' % FQDN, 'id', False)
+        capsule_features = [f['name'] for f in get_json("https://" + options.foreman_fqdn + ":" + API_PORT + "/katello/api/capsules/%s" % capsule_id)['features']]
```
evgeni: can this be an own function please?

bootstrap.py
```diff
@@ -1053,12 +1066,78 @@ def exec_service(service, command, failonerror=True):
...
+
+        # Optionally configure new hostgroup, location
+        if options.hostgroup:
+            print_running("Calling Foreman API to switch hostgroup for %s to %s" % (FQDN, options.hostgroup))
+            hostgroup_id = return_matching_foreman_key('hostgroups', 'title="%s"' % options.hostgroup, 'id', False)
+            jdata = json.loads('{"host": {"hostgroup_id": "%s"}}' % hostgroup_id)
+            put_json("https://" + options.foreman_fqdn + ":" + API_PORT + "/api/hosts/%s" % host_id, jdata)
```
evgeni: this can be replaced with `update_host_capsule_mapping()`?

bootstrap.py
```diff
@@ -1053,12 +1066,78 @@ def exec_service(service, command, failonerror=True):
...
+        if options.location:
+            print_running("Calling Foreman API to switch location for %s to %s" % (FQDN, options.location))
+            location_id = return_matching_foreman_key('locations', 'title="%s"' % options.location, 'id', False)
+            jdata = json.loads('{"host": {"location_id": "%s"}}' % location_id)
+            put_json("https://" + options.foreman_fqdn + ":" + API_PORT + "/api/hosts/%s" % host_id, jdata)
```
evgeni: this can be replaced with `update_host_capsule_mapping()`?

bootstrap.py
```diff
@@ -1053,12 +1066,78 @@ def exec_service(service, command, failonerror=True):
...
+
+        # Configure new proxy_id for Puppet (if not skipped), and OpenSCAP (if available and not skipped)
```
evgeni: can we figure out which feature were enabled on the old proxy and only try to change those? relying on the cli options seems wrong here.

bootstrap.py
```diff
@@ -1053,12 +1066,78 @@ def exec_service(service, command, failonerror=True):
...
+            # Not using clean_puppet() and install_puppet_agent() here, because
+            # that would nuke custom /etc/puppet/puppet.conf files, which might
+            # yield undesirable results.
+            print_running("Updating Puppet configuration")
+            exec_failexit("sed -i 's/^[[:space:]]*server.*/   server     = %s/' /etc/puppet/puppet.conf" % options.foreman_fqdn)
+            exec_failok("sed -i 's/^[[:space:]]*ca_server.*/   server     = %s/' /etc/puppet/puppet.conf" % options.foreman_fqdn)  # For RHEL5 stock puppet.conf
+            delete_directory("/var/lib/puppet/ssl")
+            delete_file("/var/lib/puppet/client_data/catalog/%s.json" % FQDN)
+
+            print_generic("Running Puppet in noop mode to generate SSL certs")
```
evgeni: that duplicates code from `install_puppet_agent`, while we can't use the function itself, shall we split out the "run noop and tell user to sign" part in an own function and call that it both places?

evgeni (changes requested):
some inline comments, also it's called Katello/Proxy, not Satellite/Capsule, so please update docs/option name :)

bootstrap.py
wzzrd: Suppose so. In fact, it should throw an exception if it's not.

bootstrap.py
wzzrd: I'm assuming you specifically mean the  capsule_features part? No sense in moving the whole part into a function imho.

bootstrap.py
wzzrd: Mmh. No, doesn't make sense to push it into that function, but I do think it can be it's own function that updates hostgroup or location mapping of a host.

bootstrap.py
wzzrd: Mmh. No, doesn't make sense to push it into that function, but I do think it can be it's own function that updates hostgroup or location mapping of a host.

bootstrap.py
wzzrd: This is always going to be a chicken and egg problem: even if Puppet is enabled for a system on the Foreman side, it doesn't mean the Puppet agent is actually configured and running, for example. 

For OpenSCAP I check for whether or not that's enabled on the Satellite^HForeman side.  Setting the openscap_proxy_id for a system doesn't actually do anything if no policy is assigned. If a policy *is* assigned, you want it set, if there is no policy assigned, it doesn't matter. OpenSCAP is only skipped if foreman is in options.skip, or when the new Capsule doesn't have it.

People might be running Puppet outside of the scope of Foreman, so the skip option for Puppet needs to be there. 

The Katello functionality needs always be done.

So, adding all up, I would leave the options, unless I'm overlooking something.

bootstrap.py
wzzrd: Yeah, that sounds like a plan. Will look into it.

wzzrd:
As for the Capsule / Proxy discussion: I'm not against naming everything by the upstream names, but the problem is that there is a million places where that is not the case right now. 

To be honest, looking at `get_api_port()`, that will only work on *Satellite* or *Capsule* servers, because it relies on parsing`/etc/rhsm/rhsm.conf`. That file might not even exist on pure upstream Foreman / Foreman Proxy, as I understand it. I'm not saying that is right, I'm just saying this issue needs a closer look.

bootstrap.py
wzzrd: In fact, it already does throw an error if foreman_fqdn isn't set and new_capsule is.

wzzrd:
- DRYd the Puppet message and noop run by putting them in a function
- Created a function to change hostgroup / location
- Created get_capsule_features function
- Removed superfluous check for foreman_fqdn

Please review code, but wait with potential merge until I do a final functional that I cannot do here.

wzzrd:
@evgeni ping

wzzrd:
Test done w/ RHEL7 on Satellite 6.2.12 machine and ditto Capsule. One last change incorporated, works fine otherwise: 6e8d9bee966a5dbf75463da9ea24c82a9e5f0661

evgeni:
`/etc/rhsm/rhsm.conf` exists on Katello clients, which is good enough for us :)

bootstrap.py
evgeni: yeah, I meant the get features part :)

bootstrap.py
```diff
@@ -530,9 +537,27 @@ def delete_json(url):
...
+    return put_json(url, jdata)
+
+
+def get_capsule_features(capsule_id):
+    url = "https://" + options.foreman_fqdn + ":" + str(API_PORT) + "/katello/api/capsules/%s" % str(capsule_id)
+    return [f['name'] for f in get_json(url)['features']]
+
+
+def update_host_config(attribute, value, host_id):
+    attribute_id = return_matching_foreman_key(attribute + 's', 'title="%s"' % value, 'id', False)
```
evgeni, line 631: why don't we just pass `attribute` as plural? we do it in all the other places too ;)

evgeni: ah, craps, that's because we use `_id` later on. gotcha.

wzzrd:
@evgeni @sideangleside are yall waiting for me to 's/capsule/smart_proxy/g' or is there something else?

README.md
```diff
@@ -7,6 +7,9 @@ bootstrap Script for migrating existing running systems to Foreman with the Kate
 This script can take a system that is registered to Spacewalk, Satellite 5, Red Hat
 Network Classic and get it registered to Foreman & Katello.
 
+Optionally, you can also move systems from one Satellite 6 system or Capsule to
```
evgeni: I think moving between different Satellite/Katello installs should not be done with `--new-capsule` as that expects the host to be already present in Foreman to be able to edit it.

How about:
```
Optionally, you can also move systems between Capsules (both internal and external) of one Katello installation by using the `--new-capsule` option.
```

README.md
```diff
@@ -169,6 +184,25 @@ By default, bootstrap.py does not delete the system's profile from the legacy pl
...
 
+### Migrating a system from one Satellite 6 / Capsule to another in the same infrastructure
+
+In order to manually balance the load over multiple Capsule servers, you might
+want to move some existing systems to newly deployed Capsules. You can easily
+do this by running the bootstrap.py script like the examples below. Mind that
+you still have to manually revoke any Puppet certificates on the old capsules!
+
+~~~
+# ./bootstrap.py -l admin --new-capsule  --server new.capsule.server
```
evgeni: double blank before `--server` ;)

README.md
```diff
@@ -169,6 +184,25 @@ By default, bootstrap.py does not delete the system's profile from the legacy pl
...
+~~~
+
+If you want to change the hostgroup and location of the system at the same
+time, run:
+
+~~~
+# ./bootstrap.py -l admin --new-capsule  --server new.capsule.server \
```
evgeni: double blank before `--server` ;)

bootstrap.py
```diff
@@ -1053,12 +1082,71 @@ def exec_service(service, command, failonerror=True):
         migrate_systems(options.org, options.activationkey)
         if options.enablerepos:
             enable_repos()
+    elif options.new_capsule:
```
evgeni, line 1162: if we end up here on a machine that is not registered at all yet, we get funny errors, should we check for a regged box?

wzzrd: Is there a sane and simple way to do so?

evgeni: did you just ask for a sane way inside bootstrap.py? ;)

I've been playing around with

```python
    elif options.new_capsule:
        if not os.path.exists('/etc/rhsm/ca/katello-server-ca.pem'):
            print_generic("This system is not yet registered, cannot move to a new capsule")
            sys.exit(1)
```

for now.

probably even better:

```python
        if not os.path.exists('/etc/rhsm/ca/katello-server-ca.pem') or not os.path.exists('/etc/pki/consumer/cert.pem'):
```

wzzrd: methinks that code should live in a function, too

bootstrap.py
```diff
@@ -1053,12 +1082,71 @@ def exec_service(service, command, failonerror=True):
...
+            update_host_capsule_mapping("puppet_proxy_id", capsule_id, host_id)
+            update_host_capsule_mapping("puppet_ca_proxy_id", capsule_id, host_id)
+        if 'foreman' not in options.skip and 'Openscap' in capsule_features:
+            print_running("Calling Foreman API to update OpenSCAP proxy for %s to %s" % (FQDN, options.foreman_fqdn))
+            update_host_capsule_mapping("openscap_proxy_id", capsule_id, host_id)
+        elif 'foreman' not in options.skip and 'Openscap' not in capsule_features:
+            print_warning("New capsule doesn't have OpenSCAP capability, not switching / configuring openscap_proxy_id")
+
+        print_running("Calling Foreman API to update content source for %s to %s" % (FQDN, options.foreman_fqdn))
+        update_host_capsule_mapping("content_source_id", capsule_id, host_id)
```
evgeni, line 1204: At least in Katello 3.5/3.6, this needs to be `host[content_facet_attributes][content_source_id]`, not `host[content_source_id]`

bootstrap.py
```diff
@@ -270,6 +273,16 @@ def enable_rhsmcertd():
     exec_service("rhsmcertd", "restart")
 
 
+def is_registered():
+    # Check if all required certificates are in place (i.e. a system is
+    # registered to begin with) before we start changing things
+    if (os.path.exists('/etc/rhsm/ca/katello-server-ca.pem') and
```
evgeni: I'd do just `return os.path.exists('/etc/rhsm/ca/katello-server-ca.pem') and os.path.exists('/etc/pki/consumer/cert.pem'))`

bootstrap.py
```diff
@@ -1053,12 +1092,74 @@ def exec_service(service, command, failonerror=True):
...
+        # > wil update system definition to point to new capsule for content,
+        # > Puppet, OpenSCAP and update Puppet configuration (if applicable)
+        # > MANUAL SIGNING OF CSR OR MANUALLY CREATING AUTO-SIGN RULE STILL REQUIRED!
+        # > API doesn't have a public endpoint for creating auto-sign entries yet!
+        if not is_registered():
+            print_error("This system doesn't seem to be registered to a Capsule at this moment.")
+            sys.exit(1)
+
+        # Make system ready for switch, gather required data
+        get_bootstrap_rpm(clean=True, unreg=False)
```
evgeni: I think we should call `install_prereqs()` before that. In 99.999% the cases it will be a no-op, but better safe than sorry.

wzzrd: I'll slap it in there, but if sub-man is not installed, a system cannot be registered to a Satellite, be member of a hostgroup, have a content source, etc., to begin with, right? So aren't we catering for a non-existent use case?

bootstrap.py
wzzrd: Ya, that's a nicer, more Pythonic way of doing it. Fixed.

bootstrap.py
wzzrd: Ok, @evgeni, tiny problem.

I have changed the code so that it - in case we try to change content_source_id - it creates a special jdata that has both the old `content_source_id` in there as well as the new [content_facet_attributes][content_source_id].

From a hammer point of view, this allows you to see the changed content source. From the point of view of the Satellite UI though, it doesn't. It still shows you the first content source it used, or maybe it's the server that was used to register, I don't know.

All I know is that 

    $ hammer host info --id HOSTID

will show the new content source, and the UI will not.

This is only in Satellite 6.3: in Satellite 6.2, everything works just fine.

bootstrap.py
evgeni: Yeah, it's quite a constructed case (I stumbled over it while trying to move a non-registered CentOS box before you added the "is registered" check).
The only case this can happen is: someone deliberately removed s-m *after* registering, thus mostly breaking the system.

bootstrap.py
evgeni: Mhhh. I'll have a try on monday with my Katello 3.5. I *thought* (and that's why I told you that) that this hack worked for me.

wzzrd:
@evgeni I've manually rebased on current master this morning because there were some conflicts

bootstrap.py
```diff
@@ -1101,12 +1143,75 @@ def exec_service(service, command, failonerror=True):
...
+
+        if 'puppet' not in options.skip and 'foreman' not in options.skip:
+            print_running("Stopping the Puppet agent for configuration update")
+            exec_service("puppet", "stop")
+
+            # Not using clean_puppet() and install_puppet_agent() here, because
+            # that would nuke custom /etc/puppet/puppet.conf files, which might
+            # yield undesirable results.
+            print_running("Updating Puppet configuration")
+            exec_failexit("sed -i 's/^[[:space:]]*server.*/   server     = %s/' /etc/puppet/puppet.conf" % options.foreman_fqdn)
```
evgeni: can you add puppet4/5 path support with the code introduced in https://github.com/Katello/katello-client-bootstrap/pull/230?

bootstrap.py
evgeni: yeah, can reproduce. weird.

bootstrap.py
evgeni: this seems to do it:

```diff
diff --git a/bootstrap.py b/bootstrap.py
index aaf79cc..358edf8 100644
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -1190,6 +1190,19 @@ if __name__ == '__main__':
         enable_rhsmcertd()
 
         if 'puppet' not in options.skip and 'foreman' not in options.skip:
+            puppet_major_version = get_puppet_version()
+            if puppet_major_version == 3:
+                puppet_conf_file = '/etc/puppet/puppet.conf'
+                var_dir = '/var/lib/puppet'
+                ssl_dir = '/var/lib/puppet/ssl'
+            elif puppet_major_version in [4, 5]:
+                puppet_conf_file = '/etc/puppetlabs/puppet/puppet.conf'
+                var_dir = '/opt/puppetlabs/puppet/cache'
+                ssl_dir = '/etc/puppetlabs/puppet/ssl'
+            else:
+                print_error("Unsupported puppet version")
+                sys.exit(1)
+
             print_running("Stopping the Puppet agent for configuration update")
             exec_service("puppet", "stop")
 
@@ -1197,10 +1210,10 @@ if __name__ == '__main__':
             # that would nuke custom /etc/puppet/puppet.conf files, which might
             # yield undesirable results.
             print_running("Updating Puppet configuration")
-            exec_failexit("sed -i 's/^[[:space:]]*server.*/   server     = %s/' /etc/puppet/puppet.conf" % options.foreman_fqdn)
-            exec_failok("sed -i 's/^[[:space:]]*ca_server.*/   server     = %s/' /etc/puppet/puppet.conf" % options.foreman_fqdn)  # For RHEL5 stock puppet.conf
-            delete_directory("/var/lib/puppet/ssl")
-            delete_file("/var/lib/puppet/client_data/catalog/%s.json" % FQDN)
+            exec_failexit("sed -i 's/^[[:space:]]*server.*/   server     = %s/' %s" % (options.foreman_fqdn, puppet_conf_file))
+            exec_failok("sed -i 's/^[[:space:]]*ca_server.*/   server     = %s/' %s" % (options.foreman_fqdn, puppet_conf_file))  # For RHEL5 stock puppet.conf
+            delete_directory(ssl_dir)
+            delete_file("%s/client_data/catalog/%s.json" % (var_dir, FQDN))
 
             noop_puppet_signing_run()
             print_generic("Puppet agent is not running; please start manually if required.")
```

bootstrap.py
```diff
@@ -1101,12 +1143,75 @@ def exec_service(service, command, failonerror=True):
...
+            exec_failok("sed -i 's/^[[:space:]]*ca_server.*/   server     = %s/' /etc/puppet/puppet.conf" % options.foreman_fqdn)  # For RHEL5 stock puppet.conf
```
evgeni: this should be `ca_server` ;)

evgeni:
@wzzrd puppet4 support and sed-fixes are at https://github.com/evgeni/katello-client-bootstrap/tree/capsule-migration-option-new, feel free to pull these in

wzzrd:
Pulled in latest master and changes from your fork, @evgeni

bootstrap.py
evgeni: memo to self, try to see what hammer does

evgeni: So `hammer -d host update --name centos7.kangae.example.com --content-source-id 2` shows

```
[ INFO 2018-01-22 10:55:53 API] Server: https://pipeline-katello-3-5-centos7.kangae.example.com
[ INFO 2018-01-22 10:55:53 API] PUT /api/hosts/20
[DEBUG 2018-01-22 10:55:53 API] Params: {
    "host" => {
                                 "name" => "centos7.kangae.example.com",
                   "compute_attributes" => {
            "volumes_attributes" => {}
        },
             "content_facet_attributes" => {
            "content_source_id" => "2"
        },
        "subscription_facet_attributes" => {},
           "host_parameters_attributes" => [],
                "interfaces_attributes" => []
    }
}
```

and the reply looks correct too:

```
         "content_facet_attributes" => {
                                    "id" => 18,
                                  "uuid" => "2acbdedf-46cc-4504-b0f3-e6a743f064f4",
                       "content_view_id" => 3,
                     "content_view_name" => "default-p4",
              "lifecycle_environment_id" => 1,
            "lifecycle_environment_name" => "Library",
                     "content_source_id" => 2,
                   "content_source_name" => "pipeline-proxy-3-5-centos7.kangae.example.com",
               "kickstart_repository_id" => nil,
                         "errata_counts" => {
               "security" => 0,
                 "bugfix" => 0,
            "enhancement" => 0,
                  "total" => 0
        },
              "applicable_package_count" => 0,
              "upgradable_package_count" => 0,
                          "content_view" => {
              "id" => 3,
            "name" => "default-p4"
        },
                 "lifecycle_environment" => {
              "id" => 1,
            "name" => "Library"
        },
                        "content_source" => {
              "id" => 2,
            "name" => "pipeline-proxy-3-5-centos7.kangae.example.com",
             "url" => "https://pipeline-proxy-3-5-centos7.kangae.example.com:9090"
        },
                  "content_view_version" => "1.0",
               "content_view_version_id" => 3,
                 "content_view_default?" => false,
        "lifecycle_environment_library?" => true,
               "katello_agent_installed" => true
    },
```

yet the UI still shows the wrong name

README.md
```diff
@@ -21,6 +24,18 @@ Network Classic and get it registered to Foreman & Katello.
 * Configuring the system with a proper Puppet configuration pointing at Foreman
 * Removing/disabling old RHN Classic packages/daemons (rhnsd, osad, etc)
 
+## System already registered to a Satellite 6 server / Capsule (--new-capsule)
```
evgeni: `s/Satellite 6/Katello/`

README.md
```diff
@@ -21,6 +24,18 @@ Network Classic and get it registered to Foreman & Katello.
...
+* Clean the existing Katello agent installation
+* Install the Katello consumer RPM for the target Satellite 6 server / Capsule
```
evgeni: `s/Satellite 6/Katello/`

README.md
```diff
@@ -21,6 +24,18 @@ Network Classic and get it registered to Foreman & Katello.
...
+* Install the Katello agent software again, using the configuration for the
+  target Satellite 6 / Capsule server
```
evgeni: `s/Satellite 6/Katello/`

README.md
```diff
@@ -169,6 +184,25 @@ By default, bootstrap.py does not delete the system's profile from the legacy pl
 
 There are times where it is necessary to migrate clients from one Foreman + Katello installation to another. For instance, in lieu of upgrading an older Foreman + Katello installation, you choose to build a new installation in parallel. bootstrap.py can then be used to migrate clients from one Foreman + Katello installation to another. Simply provide the `--force` option, and bootstrap.py will remove the previous `katello-ca-consumer-*` package (from the old system), and will install the `katello-ca-consumer-*` package (from the new system), and continue registration as usual.
 
+### Migrating a system from one Satellite 6 / Capsule to another in the same infrastructure
```
evgeni: `s/Satellite 6/Katello/`

README.md
```diff
@@ -169,6 +184,25 @@ By default, bootstrap.py does not delete the system's profile from the legacy pl
...
+
+In order to manually balance the load over multiple Capsule servers, you might
+want to move some existing systems to newly deployed Capsules. You can easily
+do this by running the bootstrap.py script like the examples below. Mind that
+you still have to manually revoke any Puppet certificates on the old capsules!
+
+~~~
+# ./bootstrap.py -l admin --new-capsule --server new.capsule.server
```
evgeni: I'd call the new capsule `newcapsule.example.com`

README.md
```diff
@@ -169,6 +184,25 @@ By default, bootstrap.py does not delete the system's profile from the legacy pl
...
+~~~
+
+If you want to change the hostgroup and location of the system at the same
+time, run:
+
+~~~
+# ./bootstrap.py -l admin --new-capsule --server new.capsule.server \
```
evgeni: I'd call the new capsule `newcapsule.example.com`

evgeni:
Thanks @wzzrd! I think if we update the README to say Katello and not Satellite and give the new Capsule a `example.com` domain as with the rest of the examples, this is good to go.

@sideangleside you also happy with that?

wzzrd:
Done.

sideangleside:
I _finally_ got time to review this. No issues to report from me. Other than squashing the commits, we can merge this PR. Thanks for the pretty big feature @wzzrd

wzzrd:
Squashed and pushed @sideangleside @evgeni

evgeni (approved):

PR #230, opened 2018-01-03
Make bootstrap.py compatible with Puppet 4

### LINKED ISSUE #229: Make bootstrap.py compatible with Puppet 4
Currently bootstrap.py fails if the client is trying to install puppet-agent for Puppet 4.
This seems to happen because of the change of the directory structure that is there with Puppet 4.

Puppet 3:
The configuration files live in /etc/puppet/puppet.conf
The executable is present in /usr/bin/puppet

Puppet 4:
The configuration file is in /etc/puppetlabs/puppet/puppet.conf
The executable is present in /opt/puppetlabs/puppet/bin/puppet

With the current version of bootstrap.py, it looks for the files according to Puppet 3 configuration and hence fails if the client is requesting for Puppet 4.

The question is should we detect the change in the puppet version inside bootstrap.py or allow the user to pass an option something like --install-puppet4 to handle this?

h4xr:
Fixes #229 

Signed-off-by: Saurabh Badhwar <sbadhwar@redhat.com>

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -382,6 +382,8 @@ def clean_puppet():
     print_generic("Cleaning old Puppet Agent")
     call_yum("erase", "puppet")
     delete_directory("/var/lib/puppet/")
+    delete_directory("/opt/puppetlabs/puppet/cache")
+    delete_directory("/etc/puppetlabs/puppet/ssl")
 
 
 def clean_environment():
@@ -415,18 +417,26 @@ def install_puppet_agent():
     puppet_major_version = get_puppet_version()
     if puppet_major_version == 3:
         puppet_conf_file = '/etc/puppet/puppet.conf'
+        main_section = """[main]
+vardir = /var/lib/puppet
+logdir = /var/log/puppet
+rundir = /var/run/puppet
+ssldir = $vardir/ssl
+"""
     elif puppet_major_version in [4, 5]:
         puppet_conf_file = '/etc/puppetlabs/puppet/puppet.conf'
+        main_section = """[main]
+vardir = /opt/puppetlabs/puppet/cache
+logdir = /var/log/puppetlabs/puppet
+rundir = /var/run/puppetlabs
+ssldir = /etc/puppetlabs/puppet/ssl
+"""
     else:
         print_error("Unsupported puppet version")
         sys.exit(1)
     puppet_conf = open(puppet_conf_file, 'wb')
     puppet_conf.write("""
-[main]
-vardir = /var/lib/puppet
-logdir = /var/log/puppet
-rundir = /var/run/puppet
-ssldir = $vardir/ssl
+%s
 [agent]
 pluginsync      = true
 report          = true
@@ -436,7 +446,7 @@ ca_server       = %s
 certname        = %s
 environment     = %s
 server          = %s
-""" % (options.foreman_fqdn, FQDN, puppet_env, options.foreman_fqdn))
+""" % (main_section, options.foreman_fqdn, FQDN, puppet_env, options.foreman_fqdn))
     if options.puppet_noop:
         puppet_conf.write("""noop            = true
 """)
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -383,7 +383,10 @@ def install_puppet_agent():
     call_yum("install", "puppet")
     enable_service("puppet")
 
-    puppet_conf = open('/etc/puppet/puppet.conf', 'wb')
+    if os.path.isfile('/etc/puppet/puppet.conf'):
+        puppet_conf = open('/etc/puppet/puppet.conf', 'wb')
+    else:
+        puppet_conf = open('/etc/puppetlabs/puppet/puppet.conf', 'wb')
     puppet_conf.write("""
 [main]
 vardir = /var/lib/puppet
@@ -407,7 +410,10 @@ server          = %s
     print_generic("Running Puppet in noop mode to generate SSL certs")
     print_generic("Visit the UI and approve this certificate via Infrastructure->Capsules")
     print_generic("if auto-signing is disabled")
-    exec_failexit("/usr/bin/puppet agent --test --noop --tags no_such_tag --waitforcert 10")
+    if os.path.isfile('/usr/bin/puppet'):
+        exec_failexit("/usr/bin/puppet agent --test --noop --tags no_such_tag --waitforcert 10")
+    else:
+        exec_failexit("/opt/puppetlabs/puppet/bin/puppet agent --test --noop --tags no_such_tag --waitforcert 10")
     if 'puppet-enable' not in options.skip:
         enable_service("puppet")
         exec_service("puppet", "restart")
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -218,6 +218,45 @@ def install_prereqs():
         delete_file('/etc/yum.repos.d/katello-client-bootstrap-deps.repo')
 
 
+def get_puppet_path():
+    """
+    Get the path of where the puppet excutable is located.
+    """
+    puppet_lookup_paths = [
+        "/usr/bin",
+        "/opt/puppetlabs/puppet/bin"
+    ]
+    executable_path = None
+    for path in puppet_lookup_paths:
+        executable_path = os.path.join(path, 'puppet')
+        if os.path.isfile(executable_path):
+            break
+        executable_path = None
+
+    if executable_path is None:
+        print_error("No puppet executable found")
+        sys.exit(1)
+    return executable_path
+
+
+def get_puppet_version():
+    """
+    Retrieve the Major version of puppet install.
+    """
+    puppet_executable = get_puppet_path()
+    filtered_command = filter_string("%s agent --version" % (puppet_executable))
+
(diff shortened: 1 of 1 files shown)


bootstrap.py
```diff
@@ -383,7 +383,10 @@ def install_puppet_agent():
     call_yum("install", "puppet")
     enable_service("puppet")
 
-    puppet_conf = open('/etc/puppet/puppet.conf', 'wb')
+    if os.path.isfile('/etc/puppet/puppet.conf'):
```
evgeni: can we turn that around and handle `/etc/puppetlabs` as the special case? feels slightly "cleaner" to me

bootstrap.py
```diff
@@ -407,7 +410,10 @@ def install_puppet_agent():
     print_generic("Running Puppet in noop mode to generate SSL certs")
     print_generic("Visit the UI and approve this certificate via Infrastructure->Capsules")
     print_generic("if auto-signing is disabled")
-    exec_failexit("/usr/bin/puppet agent --test --noop --tags no_such_tag --waitforcert 10")
+    if os.path.isfile('/usr/bin/puppet'):
```
evgeni: can we turn that around and test for `/opt/puppetlabs/puppet/bin/puppet`? and in the else case, run puppet without a path at all, so the shell can find it even in another folder?

evgeni (changes requested):

bootstrap.py
```diff
@@ -218,6 +218,45 @@ def install_prereqs():
...
+            break
+        executable_path = None
+
+    if executable_path is None:
+        print_error("No puppet executable found")
+        sys.exit(1)
+    return executable_path
+
+
+def get_puppet_version():
```
evgeni: I think I'd still prefer using the RPMdb like this:

```python
def get_puppet_version():
    ts = rpm.TransactionSet()
    for name in ['puppet', 'puppet-agent']:
        mi = ts.dbMatch('name', name)
        for h in mi:
            puppet_version = int(h['version'].split('.')[0])
            if h['name'] == 'puppet-agent' and puppet_version == 1:
                puppet_version = 4
            return puppet_version
```

This works fine for Puppet 3, 4 and 5

@sideangleside whatcha think?

bootstrap.py
```diff
@@ -383,7 +422,15 @@ def install_puppet_agent():
     call_yum("install", "puppet")
     enable_service("puppet")
 
-    puppet_conf = open('/etc/puppet/puppet.conf', 'wb')
+    puppet_major_version = get_puppet_version()
+    if puppet_major_version == 3:
+        puppet_conf_file = '/etc/puppet/puppet.conf'
+    elif puppet_major_version == 4:
```
evgeni: can we make this `puppet_major_version in [4,5]`? as Puppet5 uses the same path and that would make it more future proof

bootstrap.py
```diff
@@ -218,6 +218,45 @@ def install_prereqs():
         delete_file('/etc/yum.repos.d/katello-client-bootstrap-deps.repo')
 
 
+def get_puppet_path():
```
evgeni, line 221: this would be then (if we use the RPMdb)

```python
puppet_major_version = get_puppet_version()
if puppet_major_version == 3:
    return '/usr/bin/puppet'
elif puppet_major_version in [4,5]:
    return '/opt/puppetlabs/puppet/bin/puppet'
else:
    error

File: bootstrap.py (hunk 3 above)
Comment (h4xr) [reply bootstrap.py]: @evgeni If I am not getting wrong, we are comparing the puppet-agent packaging version here. I doubt if this will be future proof or not, in case packaging convention changes in future for the newer releases of Puppet.

Comment (evgeni) [reply bootstrap.py]: correct, it's using the packaging version. I would not expect the major version of the packaging to differ from the major version of the agent in the future (well, puppet4 agent is an exemption here :()

Comment (sideangleside) [reply bootstrap.py]: I am strongly supportive of checking the RPMs version number as it is consistent with other checks for packages that we do. Additionally, checking via an executable's location is just as prone to packaging changes as anything else, so i'd want to be consistent.

File: bootstrap.py (hunk 4 above)
Comment (sideangleside) [reply bootstrap.py]: 👍

**Comment by h4xr** [comment]
@evgeni Addressed the changes in the commit. Please review :)

**Comment by evgeni** [comment]
code LGTM :+1: 

I'll give it a few test runs later today and merge afterwards, unless @sideangleside is faster ;)

**Comment by evgeni** [comment]
So I gave it a fair spin, and it seems to work fine. However I spotted a few differences in the `puppet.conf` compared to the version [Foreman deploys](https://github.com/theforeman/foreman/blob/develop/app/views/unattended/provisioning_templates/snippet/_puppet.conf.erb) by default:

our:
```
vardir = /var/lib/puppet
logdir = /var/log/puppet
rundir = /var/run/puppet
ssldir = $vardir/ssl
```

foreman:
```
vardir = /opt/puppetlabs/puppet/cache
logdir = /var/log/puppetlabs/puppet
rundir = /var/run/puppetlabs
ssldir = /etc/puppetlabs/puppet/ssl
```

I think we should align this for the Puppet4/5 case.

**Comment by evgeni** [comment]
and with that change, we should also extend https://github.com/h4xr/katello-client-bootstrap/blob/2a9f0662d26b90042141a79cda9d216eb3118079/bootstrap.py#L384 to wipe `/opt/puppetlabs/puppet/cache` and `/etc/puppetlabs/puppet/ssl`

File: bootstrap.py [hunk 6]
```diff
@@ -383,13 +414,35 @@ def install_puppet_agent():
...
+""")
+    elif puppet_major_version in [4, 5]:
+        puppet_conf.write("""
+[main]
+vardir = /opt/puppetlabs/puppet/cache
+logdir = /var/log/puppetlabs/puppet
+rundir = /var/run/puppet
+ssldir = /etc/puppetlabs/puppet/ssl
+""")
+    else:
```
evgeni: that code is unreached, we already verified that it's version 3, 4 or 5 when looking for the conf path

bootstrap.py
```diff
@@ -383,13 +414,35 @@ def install_puppet_agent():
...
+    puppet_major_version = get_puppet_version()
+    if puppet_major_version == 3:
+        puppet_conf_file = '/etc/puppet/puppet.conf'
+    elif puppet_major_version in [4, 5]:
+        puppet_conf_file = '/etc/puppetlabs/puppet/puppet.conf'
+    else:
+        print_error("Unsupported puppet version")
+        sys.exit(1)
+    puppet_conf = open(puppet_conf_file, 'wb')
+    if puppet_major_version == 3:
```
evgeni: I think instead of having another set of `if/elif` here, I'd put the dirs (or the whole `[main]` section) into a variable during the conf path detection, and just use the variable in one `write()`.

evgeni (approved):
two comments, which would make the code a bit cleaner (IMHO), but neither is really blocking and the code works just fine as it is. ACK from me.

would love to hear a second ACK from @sideangleside before merging, though

bootstrap.py
```diff
@@ -383,13 +414,31 @@ def install_puppet_agent():
...
 rundir = /var/run/puppet
 ssldir = $vardir/ssl
+"""
+    elif puppet_major_version in [4, 5]:
+        puppet_conf_file = '/etc/puppetlabs/puppet/puppet.conf'
+        main_section = """
+[main]
+vardir = /opt/puppetlabs/puppet/cache
+logdir = /var/log/puppetlabs/puppet
+rundir = /var/run/puppet
```
evgeni: this should be `/var/run/puppetlabs`

bootstrap.py
```diff
@@ -383,13 +414,31 @@ def install_puppet_agent():
     call_yum("install", "puppet")
     enable_service("puppet")
 
-    puppet_conf = open('/etc/puppet/puppet.conf', 'wb')
-    puppet_conf.write("""
+    puppet_major_version = get_puppet_version()
+    if puppet_major_version == 3:
+        puppet_conf_file = '/etc/puppet/puppet.conf'
+        main_section = """
```
evgeni: cosmetic: this makes `main_section` start with a newline, and thus the generated `puppet.conf` starts with two newlines. can we change this to `main_section = """[main]…`?

bootstrap.py
```diff
@@ -383,13 +414,31 @@ def install_puppet_agent():
...
 [main]
 vardir = /var/lib/puppet
 logdir = /var/log/puppet
 rundir = /var/run/puppet
 ssldir = $vardir/ssl
+"""
+    elif puppet_major_version in [4, 5]:
+        puppet_conf_file = '/etc/puppetlabs/puppet/puppet.conf'
+        main_section = """
```
evgeni: cosmetic: this makes `main_section` start with a newline, and thus the generated `puppet.conf` starts with two newlines. can we change this to `main_section = """[main]…`?

evgeni (changes requested):
one functional, and two cosmetic comments, good to go when at least the functional is addressed.

evgeni (approved):
ACK

waiting on @sideangleside

sideangleside:
LGTM 👍

sideangleside:
Thanks for the PR @h4xr

PR #231, opened 2018-01-05
if no fqdn was detected, tell the user they can set one with --fqdn

evgeni:


sideangleside:
👍. No issues here. Rebase and merge when complete.

PR #233, opened 2018-01-05
ensure sub-man installed before installing katello ca rpm

evgeni:
the rpm requires it, but it might be not installed yet (like on plain CentOS)


sideangleside:
The `install_prereqs()` function exists to ensure that sub-man is installed and is called _before_ `get_bootstrap_rpm()` (https://github.com/Katello/katello-client-bootstrap/blob/master/bootstrap.py#L1045:L1047) 

Why do we need another call to install sub-man? (ergo, if we exit the `install_prereqs()` function without having sub-man installed, that's a bug)

evgeni:
scratch that, https://github.com/Katello/katello-client-bootstrap/pull/227 does not call `install_prereqs()`

PR #236, opened 2018-01-17
allow FIPS enabled systems to register successfully with puppet

### LINKED ISSUE #235: [RFE] Bootstrap should properly configure Puppet on a FIPS enabled system
See [bz1495215](https://bugzilla.redhat.com/show_bug.cgi?id=1495215). On FIPS enabled systems (where `/proc/sys/crypto/fips_enabled` == `1`) Bootstrap fails to run puppet (Due to puppet's default [signing algorithm](https://puppet.com/docs/puppet/5.3/configuration.html#digestalgorithm) being MD5, which isn't allowed in FIPS mode. 

Setting `digest_algorithm = sha256` in puppet.conf will allow a successful puppet run, but the signing algorithm must match on the Puppet Master.  (so this only works if both are set)

Questions/Thoughts:

- Do we default to setting the signing_algorithm to sha256 automatically when a FIPS enabled system is detected? 
- Do we make the user explicitly set 'FIPS mode' (which does the above)?
- Do we exit the script with an error?

sideangleside:
Fixes #235

### DIFF
--- a/README.md
+++ b/README.md
@@ -480,6 +480,12 @@ Options:
                         separated
 ~~~
 
+# Additional Notes
+
+## FIPS support
+
+On systems with FIPS enabled (where `/proc/sys/crypto/fips_enabled == 1`), algorithms such as MD5 are disallowed. Bootstrap will configure `digest_algorithm = sha256` in puppet.conf to allow successful puppet runs. However, the signing algorithm **must** match on the Puppet Master. It is expected that the Puppet Masters are configured with the **same** algorithm.  
+
 # Ansible integration
 
 The `bootstrap.yml` file contains a playbook for [Ansible](https://www.ansible.com/) which can be used to copy `bootstrap.py` to the target machine and execute it there with predefined parameters.
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -247,6 +247,15 @@ def get_puppet_version():
             return puppet_version
 
 
+def is_fips():
+    """
+    Checks to see if the system is FIPS enabled.
+    """
+    fips_file = open("/proc/sys/crypto/fips_enabled", "r")
+    fips_status = fips_file.read(1)
+    return fips_status == "1"
+
+
 def get_bootstrap_rpm():
     """
     Retrieve Client CA Certificate RPMs from the Satellite 6 server.
@@ -434,6 +443,9 @@ ssldir = /etc/puppetlabs/puppet/ssl
     else:
         print_error("Unsupported puppet version")
         sys.exit(1)
+    if is_fips():
+        main_section += "digest_algorithm = sha256"
+        print_generic("System is in FIPS mode. Setting digest_algorithm to SHA256 in puppet.conf")
     puppet_conf = open(puppet_conf_file, 'wb')
     puppet_conf.write("""
 %s


bootstrap.py
```diff
@@ -247,6 +247,18 @@ def get_puppet_version():
...
+def is_fips():
+    """
+    Checks to see if the system is FIPS enabled.
+    """
+    fipsfile = open("/proc/sys/crypto/fips_enabled", "r")
+    fipsStatus = fipsfile.read(1)
+    if fipsStatus == "1":
+        return True
+    else:
+        return False
```
akofink: Any reason not to do this?

```python
return fipsStatus == "1"
```

README.md
```diff
@@ -480,6 +480,16 @@ Options:
                         separated
 ~~~
 
+# Additional Notes
+
+## FIPS support
+
+On systems with FIPS enabled (where `/proc/sys/crypto/fips_enabled` == `1`), algorithms such as MD5 are disallowed.
```
akofink: I think this should be '/proc/sys/crypto/fips_enabled == 1' (but using backticks, of course)

bootstrap.py
sideangleside: It seemed more pythonic to return True/False in lieu of a value. But I can be convinced otherwise.

evgeni: `return fipsStatus == "1"` will return (True/False) just fine and is imho the pythonicst way to do it.

README.md
```diff
@@ -480,6 +480,16 @@ Options:
...
+Bootstrap will configure `digest_algorithm = sha256` in puppet.conf to allow successful puppet runs.
+However, the signing algorithm **must** match on the Puppet Master. It is expected that the Puppet Masters are configured with
+the **same** algorithm.
```
evgeni: stray linebreak

bootstrap.py
```diff
@@ -247,6 +247,18 @@ def get_puppet_version():
             return puppet_version
 
 
+def is_fips():
+    """
+    Checks to see if the system is FIPS enabled.
+    """
+    fipsfile = open("/proc/sys/crypto/fips_enabled", "r")
```
evgeni: imho this should be `fips_file`

bootstrap.py
```diff
@@ -247,6 +247,18 @@ def get_puppet_version():
...
+    fipsStatus = fipsfile.read(1)
```
evgeni: and this `fips_status`, we're in snake_land here ;)

sideangleside:
I've incorporated all of the above feedback into https://github.com/Katello/katello-client-bootstrap/pull/236/commits/97e1930388ff7bb130849fbdd724fd4e6462f8fa

bootstrap.py
```diff
@@ -247,6 +247,16 @@ def get_puppet_version():
             return puppet_version
 
 
+def is_fips():
+    """
+    Checks to see if the system is FIPS enabled.
+    """
+    fips_file = open("/proc/sys/crypto/fips_enabled", "r")
+    fips_status = fips_file.read(1)
+    if fips_status == "1":
```
evgeni: nah, just always return it.

bootstrap.py
```diff
@@ -247,6 +247,16 @@ def get_puppet_version():
             return puppet_version
 
 
+def is_fips():
+    """
+    Checks to see if the system is FIPS enabled.
+    """
+    fips_file = open("/proc/sys/crypto/fips_enabled", "r")
```
evgeni, line 254: I wonder if we should handle systems where this file does not exist. (like a CentOS container running on Debian).

bootstrap.py
sideangleside: ah. got it now.

bootstrap.py
sideangleside: Why wouldn't it exist?  (though if we are going to pretend bootstrap works on non-RHEL, we'll need a whole nother 5 pull requests)

bootstrap.py
akofink: And, for those interested, it's the Ruby way also (though I'm struggling to find a link to back myself up).

bootstrap.py
akofink: I think an EL/Fedora container running on non-EL/Fedora is the only way this could happen.

akofink (commented):
LGTM after the one `fips_status == "1"` change is updated

bootstrap.py
sideangleside: but the FIPS attribute is set at the kernel level. I suspect `/proc/sys/crypto/fips_enabled` exists on a Debian system. (Though I don't have one in front of me to confirm)

akofink: Yeah, it's dependent on the docker host. Every single OS image (centos, rhel, debian, alpine) has that file when I run docker on Fedora, but none of the images has it when I run on non-EL/Fedora, such as on OS X, for instance. To clarify, running a RHEL container does not even have that file if the container is running on an OS that doesn't have that file.

evgeni: Yeah, exactly. It's a kernel thing, so a container running on a non-EL/Fedora kernel will not have this.

It's not like I like registering containers anyways...

evgeni (approved):

akofink (approved):

PR #237, opened 2018-01-17
allow configuring timeout for API calls and subscription-manager

sideangleside:
To make bootstrap.py more tolerant to busy Katello instances, this PR:

- increases the `urllib2` timeout for connections to 900 seconds / 15 minutes
- Configures subscription-manager's `server_timeout` value to 900 seconds
- provides the end user a new option `-t` or `--timeout` to change this timeout as per their requirements.

### DIFF
--- a/README.md
+++ b/README.md
@@ -387,6 +387,24 @@ For some users who do not have a configuration management or automation solution
 
 ~~~
 
+### Changing the API/Subscription Manager timeouts
+
+On busy servers, it is sometimes useful to increase the amount of time that the system waits before timing out during registration and subscription tasks. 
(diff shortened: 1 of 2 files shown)


mccun934:
🥇 💯

evgeni (approved):
I think the only detail I don't like is the fact that we don't restore the old, original timeout setting for sub-man, but I can live with that.

sideangleside:
@evgeni On removal of the `katello-ca-consumer` package we run: `test -f /etc/rhsm/rhsm.conf.kat-backup && command cp /etc/rhsm/rhsm.conf.kat-backup /etc/rhsm/rhsm.conf` so if the user had a custom (or no) value set, it would be restored.

PR #246, opened 2018-04-02
Adds a --content-only switch

sideangleside:
This PR provides a new `--content-only` option as an alias to `--skip foreman`. This makes it more obvious what `--skip foreman` does.


fftux:
Looks pretty straightforward.  I think this will be well received.

evgeni (approved):
:crying_cat_face: but ACK

PR #250, opened 2018-04-16
support setting arbitrary servers as Puppet master and CA

evgeni:


beav:
looks good, but can you add `ca_port` as well?

evgeni:
oh dear, you can't specify `ca_server = host:port`? silly

beav:
It might, I didn't try but if it does work it would be an undocumented feature.

https://puppet.com/docs/puppet/5.4/configuration.html#caport

evgeni:
@beav updated and tested :)

sideangleside:
You'd want to add a couple of `print` statements [here](https://github.com/Katello/katello-client-bootstrap/blob/master/bootstrap.py#L1094) to output the values of the new options when `options.verbose` is true, similar to what we do for the other options. 

Also, update README.md (since you added new options, we have to show the users how to use them ). 

Otherwise, +1 on the changes. No issues on my setup.

evgeni:
@sideangleside I did as you said, not as you did.

PR #252, opened 2018-06-08
Add option for host comment in bootstrap.py

frelind:
Add option --comment for bootstrap.py


akofink (approved):
Works as expected. ACK!

PR #253, opened 2018-06-15
Add multiple REX options. See issues #204 #210 #220 (and PR220)

fcami:
--rex-capsules: Comma separated list of capsules to install Foreman's
SSH keys for remote execution. It will iterate all over the list of
capsules. If one fails user is notified by the task keeps going until
all capsules were tried.

--rex-urlkeyfile: HTTP/S location to install a file containing one or
multiple Foreman's SSH keys for remote execution. Instead of iterate
you can create file on /pub folder that contains all ssh public keys
of several capsules and tell bootstrap.py to fech it and install on
the local user authorized keys file.

--rex-authpath. Local folder where is placed the authorized_keys file
to install Foreman's SSH keys for remote execution. Default ~/.ssh.
I found that several customers specifies AuthorizedKeysFile on
sshd_config on a different path.

Closes: #204
Closes: #210
(cherry picked from commit 299b0349c3168e0cf0a5473e2a1a3e967b98a0ef)

Authored-by: Alberto Losada <alosadag@redhat.com>
Rebased-by: Evgeni Golov <evgeni@golov.de>
Rebased-by: François Cami <fcami@fedoraproject.org>

PR #254, opened 2018-07-12
Add a --force-content-source option

olardit-sg:
When creating a new host, in certain cases, we want to force the content source value to be the registration capsule.
As a matter of fact, satellite is not always setting this value, but it's needed in some remote execution configurations.
As discussed with @fcami it's usefull to workaround this bug : https://bugzilla.redhat.com/show_bug.cgi?id=1570808 

PS: I'm not respecting the too-many-statements pylint rule but I can't fix it without splitting the create_host function. As there are already some pylint exceptions in the code, let me know if you want me to add this one too, or if I need to split the function

### DIFF
--- a/README.md
+++ b/README.md
@@ -517,6 +517,8 @@ Options:
   --skip-foreman        Do not create a Foreman host. Implies --skip-puppet.
                         When using --skip-foreman, you MUST pass the
                         Organization's LABEL, not NAME
+  --force-content-source
+                        Force the content source to be the registration capsule (it overrides the value in the host group if any is defined)
   --content-only        Setup host for content only. Alias to --skip foreman.
                         Implies --skip-puppet. When using --content-only, you
                         MUST pass the Organization's LABEL, not NAME
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -735,6 +735,7 @@ def create_domain(domain, orgid, locid):
 def create_host():
 
     # pylint:disable=too-many-branches
+    # pylint:disable=too-many-statements
 
     """
     Call Foreman API to create a host entry associated with the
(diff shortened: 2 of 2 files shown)


akofink (approved):
This works as expected. ACK

sideangleside:
I am curious as to why this is needed and whether the `remote_execution_global_proxy` option has been tried. That option provides `Search for remote execution Capsule outside of the proxies assigned to the host. If locations or organizations are enabled, the search will be limited to the host's organization or location.  `

This might be useful.

olardit-sg:
Yes I tried the remote_execution_global_proxy option, but I had the following limitations:
- My locations are not fine-grained enough, so all the capsules defined in one location may not be able to reach all servers.
- If multiple capsules can potentially connect on a server, it means that I need to deploy the ssh keys of all the capsules on all the servers. For this to work I need to maintain an external list of all the capsule ssh keys and use the options of #253 . I don't think it will be very reliable in my organization. And furthermore, if I deploy a new capsule in a location I need to update all the servers of this location  with the ssh key of the new capsule. This will be a pain to maintain.

All that to say, I think it's more straightforward to use the registration capsule for remote execution as we have nothing to maintain outside of satellite. That's true that I'm loosing some resiliency as I will have only one capsule available fore remote execution, but if the capsule is down, I won't be able to get content from it too. And since, most of the use cases of remote execution will be be to install or update packages, even if remote execution works I won't be able to get the content.

fcami:
I think this is reasonable, that is, if Locations do not really match the actual deployment the best course of action is to fall back to the Content-Source capsule only, not the Satellite.

@akofink @evgeni @sideangleside do you need @olardit-sg to fix his code to make pylint happy about the too-many-statements rule, or can we make another exception?

evgeni:
I'm fine with exceptions here

akofink:
I'm also fine with an exception. We can tell pylint to ignore, right?

PR #255, opened 2018-08-02
fix(bootstrap): force lowercase FQDN writing puppet configuration

faust64:
Registering production hosts to Katello, I noticed that when an
hostname has uppercase characters, then, puppet generates its
certificate translating my FQDN to lowercases. While puppet.conf
would still mention uppercases.
Eventually, puppet agent can't fetch its catalog (403 back from
puppetserver). Updating the certname value to match our certificate
fixes.

One could say that hostnames shouldn't include uppercase characters
to begin with.
On the other hand, I can't ask my customer to rename production hosts,
without being completely certain I'm not disturbing whatever application
their servers are running.

We could try setting `--fqdn`. Although having a fool-proof puppet
configuration might still be relevant (?)
Arguably, we could also patch FQDN default value (row 1012) to
`socket.getfqdn().lower()`, although I can't guarantee this won't
break something else. AFAIU, fixing puppet.conf is just what I need for
the bootstrap to execute properly on my customers servers.

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -484,7 +484,7 @@ ca_server       = %s
 certname        = %s
 environment     = %s
 server          = %s
-""" % (main_section, options.puppet_ca_server, FQDN, puppet_env, options.puppet_server))
+""" % (main_section, options.puppet_ca_server, FQDN.lower(), puppet_env, options.puppet_server))
     if options.puppet_ca_port:
         puppet_conf.write("""ca_port         = %s
 """ % (options.puppet_ca_port))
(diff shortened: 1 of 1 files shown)


evgeni:
I agree, fixing the puppet conf only is the cleanest way to do it.

I'd ack the change if we can add a comment pointing at Puppet docs (or worst case, source) where they do/expect the `.lower`.

A quick glance says https://puppet.com/docs/puppet/5.5/configuration.html#certname is what we want:
```
For best compatibility, you should limit the value of certname to only use lowercase letters, numbers, periods, underscores, and dashes. (That is, it should match /A[a-z0-9._-]+Z/.)
```

evgeni:
LGTM

@sideangleside what do you think?

sideangleside:
:+1:

evgeni (approved):

PR #257, opened 2018-08-07
improve the error message when a search fails

evgeni:


bootstrap.py
```diff
@@ -699,7 +699,8 @@ def return_matching_key(api_path, search_key, return_key, null_result_ok=False):
     elif result_len == 0 and null_result_ok is True:
         return None
     else:
-        print_error("%d element in array for search key '%s' in API '%s'. Please note that all searches are case-sensitive. Fatal error." % (result_len, search_key, api_path))
+        print_error("%d element in array for search key '%s' in API '%s'. Please note that all searches are case-sensitive." % (result_len, search_key, api_path))
+        print_error("Please also ensure the user can see the searched objects. Fatal error.")
```
sideangleside: I would change this to `Please also ensure that the user has permissions to view the search objects`

Otherwise, ack.

sideangleside (changes requested):

bootstrap.py
evgeni: yeah.

sideangleside (approved):

PR #258, opened 2018-08-07
default bootstrap.py to install katello-host-tools and omit the katello-agent package

### LINKED ISSUE #251: [RFE] update bootstrap.py to only install katello-host-tools
Katello [docs](https://theforeman.org/plugins/katello/nightly/installation/clients.html) recommend *not* installing katello-agent. bootstrap should do the same.

sideangleside:
Fixes #251


bootstrap.py
```diff
@@ -1334,7 +1335,7 @@ def exec_service(service, command, failonerror=True):
         # >                  optionally clean and install Puppet agent
         # >                  optionally remove legacy RHN packages
         if 'katello-agent' not in options.skip:
```
evgeni: The skip option should not be `katello-agent` now anymore, right? :)

bootstrap.py
```diff
@@ -397,12 +397,13 @@ def clean_katello_agent():
     delete_file("/etc/rhsm/ca/katello-server-ca.pem")
 
 
-def install_katello_agent():
```
evgeni: Given `katello-agent` is still required for errata installation without REX, I wonder if we should leave that function in and add some `--install-katello-agent` swtich.

Just installing the package is not enough in the case you're migrating from one Katello install to another, as the pkg might be already installed (→ NOOP) but the daemon is not restarted for the new config.

evgeni (changes requested):

bootstrap.py
sideangleside: Yeah, I debated this. Changing this breaks a working config (`--skip katello-agent` now doesn't work).

bootstrap.py
sideangleside: To migrate from one Katello install to another, you must call bootstrap.py with `--force`, which also means the `clean_katello_agent` function gets called. See https://github.com/Katello/katello-client-bootstrap/blob/master/bootstrap.py#L1328 and https://github.com/Katello/katello-client-bootstrap/blob/master/bootstrap.py#L259:L266

bootstrap.py
evgeni: I'd say leave `katello-agent` in the skip list, but make it basically a NOOP, as this is the default now. 

And add `k-h-t` as a new skip item. Or maybe don't allow skipping it at all?

bootstrap.py
evgeni: True. Why did we have the restart in there in the first place then? :)

bootstrap.py
sideangleside: If we don't allow skipping `k-h-t` then we'd just get a PR for it in a few weeks anyway :)

bootstrap.py
sideangleside: I seem to remember that some older builds of `katello-agent` neither (re)started nor enabled the `goferd` daemon

bootstrap.py
evgeni: @sideangleside circling back to this. I still think skipping `katello-agent` should be a NOOP, and `k-h-t` should get an own skip step.

after that, I am good with merging this

bootstrap.py
```diff
@@ -1377,7 +1384,9 @@ def exec_service(service, command, failonerror=True):
             if hostid is not None:
                 disassociate_host(hostid)
                 delete_host(hostid)
-        if 'katello-agent' not in options.skip:
+        if 'katello-agent' in options.skip:
+            print_warning("Skipping the installation of the Katello Agent is now the default behavior. passing --skip katello-agent is deprecated")
+        if 'katello-agent' or 'katello-host-tools' not in options.skip:
```
evgeni: this is always `True`, you want:

```suggestion
        if 'katello-agent' not in options.skip or 'katello-host-tools' not in options.skip:
```

evgeni (changes requested):
one nitpick, then good to go

bootstrap.py
sideangleside: :+1: perfecto.

evgeni (approved):

PR #261, opened 2018-08-22
add option to ignore registration failures

### LINKED ISSUE #260: Allow bootstrap.py to optionally ignore subscription-manager error when there are no subscriptions allocated.
There are scenarios in which it is okay for the user to register a client to satellite without having a subscription allocated to it.  Right now, bootstrap.py fails and exits since it uses  `exec_failexit` on `subscription-manager register` call as shown below.

```
def register_systems(org_name, activationkey):
    """
    Register the host to Satellite 6's organization using
    `subscription-manager` and the given activation key.
    Option "--force" is given further.
    """
    ...
    ...
    ...
    exec_failexit("/usr/sbin/subscription-manager register --org '%s' --name '%s' --activationkey '%s' %s" % (org_label, FQDN, activationkey, options.smargs))
```
the above-mentioned line should use `exec_failok` if the user wants to optionally ignore non-availability of subscriptions in certain scenarios.

sideangleside:
Fixes #260

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -348,8 +348,12 @@ def migrate_systems(org_name, activationkey):
     else:
         options.rhsmargs += " --keep"
     exec_failok("/usr/sbin/subscription-manager config --server.server_timeout=%s" % options.timeout)
-    exec_failexit("/usr/sbin/rhn-migrate-classic-to-rhsm --org %s --activation-key '%s' %s" % (org_label, activationkey, options.rhsmargs))
-    exec_failexit("subscription-manager config --rhsm.baseurl=https://%s/pulp/repos" % options.foreman_fqdn)
+    if options.ignore_registration_failures:
+        exec_failok("/usr/sbin/rhn-migrate-classic-to-rhsm --org %s --activation-key '%s' %s" % (org_label, activationkey, options.rhsmargs))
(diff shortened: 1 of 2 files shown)


bootstrap.py
```diff
@@ -348,8 +348,12 @@ def migrate_systems(org_name, activationkey):
     else:
         options.rhsmargs += " --keep"
     exec_failok("/usr/sbin/subscription-manager config --server.server_timeout=%s" % options.timeout)
-    exec_failexit("/usr/sbin/rhn-migrate-classic-to-rhsm --org %s --activation-key '%s' %s" % (org_label, activationkey, options.rhsmargs))
-    exec_failexit("subscription-manager config --rhsm.baseurl=https://%s/pulp/repos" % options.foreman_fqdn)
+    if options.ignore_registration_failures:
```
evgeni, line 351: reading these branches, I wonder how often we will forget to update params in one of them.

how about we mangle `exec_failok` and `exec_failexit` into *one* function `exec(command, failok=False)` and make the other two sensible aliases to it? then we can just pass `ignore_r_f` as the `failok` param.

sideangleside: ack, but I'd rather handle a refactoring of the `exec*` functions as a different PR. (because I am bound to mess it up :) )

evgeni (approved):

PR #262, opened 2018-08-30
set correct entries for Puppet Master and CA entries during execution

snagoor:
With https://github.com/Katello/katello-client-bootstrap/pull/250, we now have the ability to set the arbitrary  Puppet Master and CA options, but the current codebase sets both these values to `smart_proxy_id`, which in turn defaults to `foreman_fqdn`. 

With this new change, added a couple of new entries to pull the right Puppet Master and CA Smart proxy IDs and set them accordingly.

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -1421,6 +1421,8 @@ if __name__ == '__main__':
         install_katello_agent()
         API_PORT = get_api_port()
         smart_proxy_id = return_matching_foreman_key('smart_proxies', 'name="%s"' % options.foreman_fqdn, 'id', False)
+        puppet_master_smart_proxy_id = return_matching_foreman_key('smart_proxies', 'name="%s"' % options.puppet_server, 'id', True)
+        puppet_ca_smart_proxy_id = return_matching_foreman_key('smart_proxies', 'name="%s"' % options.puppet_ca_server, 'id', True)
         current_host_id = return_matching_foreman_key('hosts', 'name="%s"' % FQDN, 'id', False)
         capsule_features = get_capsule_features(smart_proxy_id)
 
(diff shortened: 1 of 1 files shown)


snagoor:
@evgeni can you help me with this PR

evgeni:
@snagoor what kind of help do you need here? the code looks sane on the first glance.

snagoor:
@evgeni, Sorry for the confusion. What I meant was to review this code change at your leisure and let me know if there are any changes required.

evgeni:
@snagoor ah :)

So the code itself looks correct. We could argue if we want to "optimize" the runtime by only querying the IDs if the hostnames are different to `foreman_fqdn`, but I don't really think it's worth the effort.

However, I fear there is a conceptual problem here. You expect that `puppet_server` and `puppet_ca_server` exist in Foreman as Smart Proxies. But there is really no requirement for that. The user might very well run an external Puppet Server, or use a load-balanced hostname for it as we describe in [Load-Balanced Smart Proxies with HAProxy](https://theforeman.org/2018/05/load-balanced-smart-proxies-with-haproxy.html). In both cases, you will get no results from the API.

What we could do is to call `return_matching_foreman_key` with `null_result_ok=True` and only call `update_host_capsule_mapping` if we actually got an `id` and not `None`.

What do you think?

evgeni (changes requested):
see comment above

evgeni:
@snagoor ping? :)

snagoor:
> @snagoor ping? :)

Sorry, @evgeni for the delay. I will work on your suggestion

bootstrap.py
```diff
@@ -1276,9 +1278,10 @@ def exec_service(service, command, failonerror=True):
 
         # Configure new proxy_id for Puppet (if not skipped), and OpenSCAP (if available and not skipped)
         if 'foreman' not in options.skip and 'puppet' not in options.skip:
-            print_running("Calling Foreman API to update Puppet master and Puppet CA for %s to %s" % (FQDN, options.foreman_fqdn))
-            update_host_capsule_mapping("puppet_proxy_id", smart_proxy_id, current_host_id)
-            update_host_capsule_mapping("puppet_ca_proxy_id", smart_proxy_id, current_host_id)
+            print_running("Calling Foreman API to update Puppet Master for %s to %s" % (FQDN, options.puppet_server))
```
evgeni: these lines should be giarded by `if puppet_master_smart_proxy_id` as `return_matching_foreman_key` will return `None` if no matching proxy is found.

bootstrap.py
```diff
@@ -1276,9 +1278,10 @@ def exec_service(service, command, failonerror=True):
...
+            update_host_capsule_mapping("puppet_proxy_id", puppet_master_smart_proxy_id, current_host_id)
+            print_running("Calling Foreman API to update Puppet CA for %s to %s" % (FQDN, options.puppet_ca_server))
```
evgeni: these lines should be giarded by `if puppet_ca_smart_proxy_id` as `return_matching_foreman_key` will return `None` if no matching proxy is found.

evgeni (changes requested):

bootstrap.py
snagoor: @evgeni modified as per your suggestion. Please take a look at your leisure.

bootstrap.py
```diff
@@ -1434,9 +1436,15 @@ def exec_service(service, command, failonerror=True):
 
         # Configure new proxy_id for Puppet (if not skipped), and OpenSCAP (if available and not skipped)
         if 'foreman' not in options.skip and 'puppet' not in options.skip:
-            print_running("Calling Foreman API to update Puppet master and Puppet CA for %s to %s" % (FQDN, options.foreman_fqdn))
-            update_host_capsule_mapping("puppet_proxy_id", smart_proxy_id, current_host_id)
-            update_host_capsule_mapping("puppet_ca_proxy_id", smart_proxy_id, current_host_id)
+            if puppet_master_smart_proxy_id is not None and puppet_ca_smart_proxy_id is not None:
```
evgeni, line 1439: you can write this without the `is not None` part, which I find better readable
```suggestion
            if puppet_master_smart_proxy_id and puppet_ca_smart_proxy_id:
```

bootstrap.py
```diff
@@ -1434,9 +1436,15 @@ def exec_service(service, command, failonerror=True):
...
+                print_running("Calling Foreman API to update Puppet Master for %s to %s" % (FQDN, options.puppet_server))
+                update_host_capsule_mapping("puppet_proxy_id", puppet_master_smart_proxy_id, current_host_id)
+                print_running("Calling Foreman API to update Puppet CA for %s to %s" % (FQDN, options.puppet_ca_server))
+                update_host_capsule_mapping("puppet_ca_proxy_id", puppet_ca_smart_proxy_id, current_host_id)
+            else:
```
evgeni, line 1444: I wonder if we need this branching here. The `if/else` is only here to generate slightly different output, right? I think the user won't lose much if we drop the `to %s` part of the messages and then we can drop the whole `if/else` completely if we set `puppet_master_smart_proxy_id` and `puppet_ca_smart_proxy_id` to `smart_proxy_id` if the search didn't return anything.

Now, after writing this, I wonder if we need to set these to the Smart Proxy at all if the user asked for an external (not-found) Puppet setup. Because if the user bootstraps a machine against `foreman.example.com`, but is using `puppet.example.com` for Puppet Master and CA, why would we set the Puppet options in Foreman to `foreman.example.com`? That'd be wrong, no?

@sideangleside thoughts?

PR #274, opened 2018-09-20
don't fail if yum could not remove packages

evgeni:
also use yum remove everywhere, not both remove and erase


ehelms:
Needs a rebase

evgeni:
@ehelms rebased

ehelms (approved):

PR #281, opened 2018-12-18
allow defining where to get ssh keys and where to store them

evgeni:
--rex-proxies: Comma separated list of Proxies to install Foreman's
SSH keys for remote execution. It will iterate all over the list of
capsules. If one fails user is notified by the task keeps going until
all proxies were tried.

--rex-urlkeyfile: HTTP/S location to install a file containing one or
multiple Foreman's SSH keys for remote execution. Instead of iterate
you can create file on /pub folder that contains all ssh public keys
of several capsules and tell bootstrap.py to fech it and install on
the local user authorized keys file.

--rex-authpath. Local folder where is placed the authorized_keys file
to install Foreman's SSH keys for remote execution. Default ~/.ssh.
I found that several customers specifies AuthorizedKeysFile on
sshd_config on a different path.

Closes: #204
Closes: #210 

Authored-by: Alberto Losada <alosadag@redhat.com>
Rebased-by: Evgeni Golov <evgeni@golov.de>
Rebased-by: François Cami <fcami@fedoraproject.org>

### DIFF
--- a/README.md
+++ b/README.md
@@ -295,6 +295,46 @@ cat ~/.ssh/authorized_keys
 ssh-rsa AAAAB3Nz.... foreman-proxy@foreman.example.com
 ~~~
 
+#### Fetching Remote Execution SSH keys from an URL
+
+~~~
+# ./bootstrap.py -l admin \
+    -s foreman.example.com \
+    -o "Red Hat" \
+    -L RDU \
+    -g "RHEL7/Crash" \
+    -a ak-Reg_To_Crash \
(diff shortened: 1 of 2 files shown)


fcami:
LGTM @evgeni

sideangleside (approved):
I'd add an example use-case in README.md like we do for all new options. Otherwise, this LGTM.

evgeni:
added examples, merging when tests pass

PR #282, opened 2018-12-18
allow fetching REX keys from the API

evgeni:
this requires REX 1.6.7 or newer to be installed

includes #281 

so far untested :)

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -643,22 +643,10 @@ def fully_update_the_box():
 
 # curl https://satellite.example.com:9090/ssh/pubkey >> ~/.ssh/authorized_keys
 # sort -u ~/.ssh/authorized_keys
-def install_foreman_ssh_key(remote_url):
+def install_ssh_key_from_url(remote_url):
     """
-    Download and install the Satellite's SSH public key into the foreman user's
-    authorized keys file location, so that remote execution becomes possible.
-    If not set default is ~/.ssh/authorized_keys
+    Download and install Foreman's SSH public key.
     """
-    if not options.remote_exec_authpath:
-        userpw = pwd.getpwnam(options.remote_exec_user)
-        options.remote_exec_authpath = os.path.join(userpw.pw_dir, '.ssh', 'authorized_keys')
-        foreman_ssh_dir = os.path.join(userpw.pw_dir, '.ssh')
-        if not os.path.isdir(foreman_ssh_dir):
-            os.mkdir(foreman_ssh_dir, OWNER_ONLY_DIR)
-            os.chown(foreman_ssh_dir, userpw.pw_uid, userpw.pw_gid)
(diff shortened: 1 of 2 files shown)


fcami:
Not tested either, but that's the right way to do it. LGTM!

bootstrap.py
```diff
@@ -1383,4 +1410,15 @@ def exec_service(service, command, failonerror=True):
             remove_obsolete_packages()
 
         if options.remote_exec:
-            install_foreman_ssh_key()
+            if options.remote_exec_proxies:
+                listproxies = options.remote_exec_proxies.split(",")
+                for proxy_fqdn in listproxies:
+                    remote_exec_url = "https://" + str(proxy_fqdn) + ":9090/ssh/pubkey"
```
ehelms, line 1519: Not great if we ever move around the smart proxy port or someone configures it differently. Fine to keep, worth noting.

bootstrap.py
```diff
@@ -587,14 +580,44 @@ def install_foreman_ssh_key():
...
-        if foreman_ssh_key in open(foreman_ssh_authfile, 'r').read():
-            print_generic("Foreman's SSH key is already present in %s" % foreman_ssh_authfile)
+    install_ssh_key_from_string(foreman_ssh_key)
+
+
+def install_ssh_key_from_api():
+    """
+    Download and install all Foreman's SSH public keys.
+    """
+    url = "https://" + options.foreman_fqdn + ":" + str(API_PORT) + "/api/v2/smart_proxies/"
```
ehelms, line 672: This could use some apypie love (if it wasn't designed as a stand-alone script to install).

ehelms (approved):

bootstrap.py
evgeni: true, will add to the longish "refactor&improve" list :)

bootstrap.py
evgeni: bootstrap.py has to run on python with only stdlib available :/

sideangleside (approved):
A couple of questions:

- if this requires a specific version of REX, is there any way to check for that version and fail gracefully if not? 
- Doesn't the API require auth for most things? so wouldn't you have to ensure that the user has authenticated when using this capability?

evgeni:
> if this requires a specific version of REX, is there any way to check for that version and fail gracefully if not?

I can look into this. But the code will fail gracefully anyways, as in older versions there would be just no key listed, and then we don't add it. But we also don't tell the user that we didn't.

> Doesn't the API require auth for most things? so wouldn't you have to ensure that the user has authenticated when using this capability?

Correct, this only works if you pass user/password. Want me to add a check for that?

ehelms:
Needs an ole rebase!

evgeni:
rebased

Issue #171, opened 2017-03-22
Katello bootstrap.py fails to register existing FreeIPA client


ehelms:
Moved from http://projects.theforeman.org/issues/18938

```
When bootstrapping an existing FreeIPA client system will fail to "host-add" client to REALM. The --force option does not ignore the error and bootstrap will abort midway through process.

Intended operation should ignore the error when using the --force option or unregster / re-register system.
```

PR #288, opened 2019-01-28
don't use json.loads to generate dicts

evgeni:
no idea why we did that to begin with, the result is a dict in either case, and this is more readable

this includes #287 because I'm too lazy to rebase it after #287 gets merged


sideangleside:
Does passing `build=false` also prevent orchestration of the other proxy features (such as DNS, DHCP, or REALM)? If yes, this would address #171 as well.

evgeni:
> Does passing `build=false` also prevent orchestration of the other proxy features (such as DNS, DHCP, or REALM)? If yes, this would address #171 as well.

Most probably, yes. I have no setup to test that at the moment :(

Also, wrong PR ;)

ehelms (approved):

PR #289, opened 2019-02-07
don't fail if yum clean fails

evgeni:
this might happen sometimes, but it's not important enough to make
bootstrap bail


sideangleside:
Ack: :+1:

PR #290, opened 2019-02-12
Add a --set-auto-attach option

olardit-sg:
In some cases we want to set the content-host's auto-attach value to false (or true).
This is different from the auto-attach value used in the activation key(s) which
is only used at registration time. If some auto-attach actions are triggered later on
it will use the content-host's auto-attach value

For example, with a satellite in "Organization Environment Access" we typically have no subscription attached to the content-host (auto-attach = false and no sub on the AK), but if the content-host's auto-attach is set to true, at the next rhsmcertd run on the client, it will pick a subscription. 
This as been discussed with the support in https://access.redhat.com/support/cases/#/case/02196634 and in bugzilla https://bugzilla.redhat.com/show_bug.cgi?id=1676554 

NB: I have added some pylint exclusions for distutils.util module. The module is in python since at least 2.4.3 (tested on RH5.3) and pylint on my system is also happy with this import ...

### DIFF
--- a/README.md
+++ b/README.md
@@ -597,6 +597,8 @@ Options:
   --legacy-purge        Purge system from the Legacy environment (e.g. Sat5)
   -a ACTIVATIONKEY, --activationkey=ACTIVATIONKEY
                         Activation Key to register the system
+  --set-auto-attach     Set the content-host's auto-attach value to true or
+                        false
   -P, --skip-puppet     Do not install Puppet
   --skip-foreman        Do not create a Foreman host. Implies --skip-puppet.
                         When using --skip-foreman, you MUST pass the
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -47,6 +47,7 @@ import pwd
 import glob
 import shutil
 import tempfile
+import distutils.util  # pylint:disable=import-error,no-name-in-module
 from datetime import datetime
 from optparse import OptionParser
 try:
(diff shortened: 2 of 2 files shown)


sideangleside:
Firstly, thanks for the PR, as an FYI, [bz1591383](https://bugzilla.redhat.com/show_bug.cgi?id=1591383) will (once it lands in a build) disable auto-attach globally when the contentAccessMode is set to `org_environment`, making the need to set this explicitly on the client no longer required. 

Also, while your patch works, it requires authentication (to use the API), so it wouldn't work when `--skip foreman` or `--content-only` are passed. 

As an alternative, you can pass `subscription-manager auto-attach --disable` post registration to achieve the same effect.

olardit-sg:
Hello,
[bz1591383](https://bugzilla.redhat.com/show_bug.cgi?id=1591383) is a good news, it will definitely help!
Regarding `subscription-manager auto-attach --disable` I first use this, but in this case I found out that the auto-attach flag value, even if effective on the host, was not reflecting on the satellite side. That's why I switched to doing an API call to have consistency. What about using subscription-manager only if 
`--skip foreman` or `--content-only` are passed?
Also I don't know which of  a new bootstrap.py version or a candlepin version will be incorporated first in satellite, so I don't know if it makes sense to go further with this.
Let me know your thoughts

wbclark:
@olardit-sg thank you for the contribution. Since the fix for [bz1591383](https://bugzilla.redhat.com/show_bug.cgi?id=1591383) is available with Satellite 6.6 and later, we'll close this PR as no longer needed. However if we are overlooking any additional use case for this feature which is not addressed by the above BZ, please feel free to reach out.

PR #295, opened 2019-02-16
[WIP] Connection Healthcheck option

sideangleside:
Based upon a few requests I have gotten, I want to propose the following patch to check the health of a Foreman installation prior to registration. The general premise being:

- connect to https://FOREMAN:API_PORT/katello/api/ping
- Check the services to ensure that
  - all of them report *OK* as their status
  - they are not too _busy_ (defined by a user-configurable value)
 
If the services fail the health-check, exit, and remove the `katello-ca-consumer-latest` package. 

Questions that I have:

- is a response time of 1000 milliseconds an appropriate amount of time to consider a service too busy to service clients
- should we finish up #209 in this PR? (In that PR, we were attempting to validate creds _before_ installing the `katello-ca-consumer-latest` RPM, which I don't think we need to do. (assuming we clean up if the creds fail).

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -1013,6 +1013,27 @@ def check_rhn_registration():
     return False
 
 
+def validate_connection():
+    """Validate that servers are up """
+    print_generic("Beginning service status and response-time check on port %s " % API_PORT)
+    print_generic("Ensuring services are online and accessible on within %s milliseconds" % options.service_timeout)
+    url = "https://" + options.foreman_fqdn + ":" + API_PORT + "/katello/api/ping"
+    features = get_json(url)['services']
+    svc_list = ["candlepin", "pulp", "pulp_auth", "candlepin_auth", "foreman_tasks"]
+    for service in svc_list:
+        if features[service]['status'] != 'ok':
+            print_error("Error detected in the Service Health Check. Cleaning up install")
+            clean_katello_agent()
+            print_error("service %s reports a state of %s" % (service, features[service]['status']))
+            sys.exit(3)
+        if int(features[service]['duration_ms']) >= options.service_timeout:
+            print_error("Error detected in the Service Health Check. Cleaning up install")
+            clean_katello_agent()
+            print_error("service %s reponds in %s ms which exceeds maximum of %s ms" % (service, features[service]['duration_ms'], str(options.service_timeout)))
+            sys.exit(3)
+    print_success("Service status and response-time check finished")
+
+
 def enable_repos():
     """Enable necessary repositories using subscription-manager."""
     repostoenable = " ".join(['--enable=%s' % i for i in options.enablerepos.split(',')])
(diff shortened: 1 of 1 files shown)


bootstrap.py
```diff
@@ -1013,6 +1013,27 @@ def check_rhn_registration():
     return False
 
 
+def validate_connection():
+    """Validate that servers are up """
+    print_generic("Beginning service status and response-time check on port %s " % API_PORT)
+    print_generic("Ensuring services are online and accessible on within %s milliseconds" % options.service_timeout)
+    url = "https://" + options.foreman_fqdn + ":" + API_PORT + "/katello/api/ping"
+    features = get_json(url)['services']
+    svc_list = ["candlepin", "pulp", "pulp_auth", "candlepin_auth", "foreman_tasks"]
```
evgeni, line 1022: this list will vary between Katello versions, why not iterating over `features.keys()` directly?

bootstrap.py
```diff
@@ -1339,6 +1361,7 @@ def exec_service(service, command, failonerror=True):
         print("PUPPET CA PORT - %s" % options.puppet_ca_port)
         print("IGNORE REGISTRATION FAILURES - %s" % options.ignore_registration_failures)
         print("PRESERVE RHSM PROXY CONFIGURATION - %s" % options.preserve_rhsm_proxy)
+        print("SERVICE TIMEOUT - %s" % options.service_timeout)
```
evgeni, line 1364: *KATELLO* SERVICE TIMEOUT

bootstrap.py
```diff
@@ -1013,6 +1013,27 @@ def check_rhn_registration():
...
 
 
+def validate_connection():
+    """Validate that servers are up """
+    print_generic("Beginning service status and response-time check on port %s " % API_PORT)
+    print_generic("Ensuring services are online and accessible on within %s milliseconds" % options.service_timeout)
+    url = "https://" + options.foreman_fqdn + ":" + API_PORT + "/katello/api/ping"
+    features = get_json(url)['services']
+    svc_list = ["candlepin", "pulp", "pulp_auth", "candlepin_auth", "foreman_tasks"]
+    for service in svc_list:
```
evgeni, line 1023: the next 10 lines are pretty repetitive, how about:

```python
    for service in svc_list:
        service_error = None
        if features[service]['status'] != 'ok':
            service_error = "service %s reports a state of %s" % (service, features[service]['status'])
        elif int(features[service]['duration_ms']) >= options.service_timeout:
            service_error = "service %s reponds in %s ms which exceeds maximum of %s ms" % (service, features[service]['duration_ms'], str(options.service_timeout))
        if service_error:
            print_error("Error detected in the Service Health Check. Cleaning up install")
            clean_katello_agent()
            print_error(service_error)
            sys.exit(3)
```

evgeni:
I *think* 1000ms is enough, but I've long not seen heavy used katello installs :woman_shrugging: 

One thing that bothers me with this setup: we do changes to the bootstraped system, before telling "oh, no, we won't bootstrap today". And while we do try to revert those changes (you clean the ca rpm), the system is not (might not) be bit-identical as before. So yes, I think piggy-backing this ontop of #209 and making it not even install the CA RPM would be awesome.

(My worry case is when migrating from one Katello to another, we already removed the *old* CA rpm, but we cannot reliably restore it and thus leave the machine without updates)

Issue #302, opened 2019-03-27
RHEL 8 requires /usr/libexec/platform-python


swadeley:
Hello

On a RHEL8 Beta host you can run bootstrap.py as follows:

~#] /usr/libexec/platform-python bootstrap.py

or edit the hash bang line to be  #!/usr/libexec/platform-python

Should be just add this info to the built in docs?

Thank you

PR #303, opened 2019-03-27
Use platform-python on RHEL8 (#302)

swadeley:
/usr/libexec/platform-python bootstrap.py is the recommend way on RHEL8

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,7 @@ When using the `--legacy-purge` option, a user account on the legacy environment
 
 # Usage:
 
-On a RHEL8 host, call the bootstrap.py script as follows:
+On an EL8 (RHEL8, CentOS8, etc) host, there is no `/usr/bin/python` or `/usr/bin/python3` by default. The `bootstrap.py` script can be used with the `platform-python` as follows:
 ~~~
 # /usr/libexec/platform-python bootstrap.py
 ~~~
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -113,7 +113,12 @@ When using the `--legacy-purge` option, a user account on the legacy environment
 * system group administrator for a system group that the system is a member of
 * granted permissions to the system explicitly via Users-> account-> 'Systems Administered by this User'
 
-# Usages:
+# Usage:
+
+On a RHEL8 host, call the bootstrap.py script as follows:
```
evgeni:
```suggestion
On an EL8 (RHEL8, CentOS8, etc) host, there is no `/usr/bin/python` or `/usr/bin/python3` by default. The `bootstrap.py` script can be used with the `platform-python` as follows:
```

bootstrap.py
```diff
@@ -5,6 +5,7 @@
 
 Use `pydoc ./bootstrap.py` to get the documentation.
 Use `awk -F'# >' 'NF>1 {print $2}' ./bootstrap.py` to see the flow.
+Use `/usr/libexec/platform-python bootstrap.py` on RHEL8
```
evgeni, line 8: Is that important here?

evgeni (changes requested):

bootstrap.py
swadeley: no

swadeley: I think you can just merge and loose that or you prefer I force push to remove it.?

evgeni: Nah, I'm good.

evgeni (approved):

PR #304, opened 2019-04-01
BZ #1693136 - always add a newline before writing the REX key

evgeni:
also strip any possibly present newlines from the key before writing

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -692,6 +692,7 @@ def install_ssh_key_from_string(foreman_ssh_key):
     If not set default is ~/.ssh/authorized_keys
     """
     print_generic("Installing Remote Execution SSH key for user %s" % options.remote_exec_user)
+    foreman_ssh_key = foreman_ssh_key.strip()
     userpw = pwd.getpwnam(options.remote_exec_user)
     if not options.remote_exec_authpath:
         options.remote_exec_authpath = os.path.join(userpw.pw_dir, '.ssh', 'authorized_keys')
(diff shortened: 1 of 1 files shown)


ehelms (approved):

bootstrap.py
```diff
@@ -707,6 +708,7 @@ def install_ssh_key_from_string(foreman_ssh_key):
             print_generic("Foreman's SSH key already present in %s" % options.remote_exec_authpath)
             return
     output = os.fdopen(os.open(options.remote_exec_authpath, os.O_WRONLY | os.O_CREAT, OWNER_ONLY_FILE), 'a')
+    output.write("\n")
     output.write(foreman_ssh_key)
```
ekohl, line 712: Would this now need a newline at the end?

evgeni: Nah, the file will just not have an ending `\n`, which I think is fine.

ekohl: I think I've seen some tools fail on that so perhaps better to be safe than sorry?

evgeni: OpenSSH is fine with this. And adding a newline here would technically add two newslines between each key we added. Or we'd need to read the whole file, `strip()` it, and then add stuff, which I'd like to avoid.

ekohl: It's already likely the user has a line ending so you might be injecting additional newlines anyway. But I'm fine either way.

ekohl (approved):

PR #307, opened 2019-04-15
only execute API calls of "new capsule" mode when Foreman is not skipped

evgeni:
otherwise we end up with errors when people run bootstrap with `--skip foreman --new-capsule` in multihomed environments

this also makes the puppet-reconfiguration not dependent on foreman

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -1445,6 +1445,8 @@ if __name__ == '__main__':
             install_katello_agent()
         if 'katello-agent' in options.skip:
             print_warning("Skipping the installation of the Katello Agent is now the default behavior. passing --skip katello-agent is deprecated")
+        enable_rhsmcertd()
+
         API_PORT = get_api_port()
         if 'foreman' not in options.skip:
             current_host_id = return_matching_foreman_key('hosts', 'name="%s"' % FQDN, 'id', False)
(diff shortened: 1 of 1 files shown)


bootstrap.py
```diff
@@ -1445,36 +1445,37 @@ def exec_service(service, command, failonerror=True):
...
+                update_host_capsule_mapping("openscap_proxy_id", smart_proxy_id, current_host_id)
+            else:
+                print_warning("New capsule doesn't have OpenSCAP capability, not switching / configuring openscap_proxy_id")
 
-        enable_rhsmcertd()
+            print_running("Calling Foreman API to update content source for %s to %s" % (FQDN, options.foreman_fqdn))
+            update_host_capsule_mapping("content_source_id", smart_proxy_id, current_host_id)
 
-        if 'puppet' not in options.skip and 'foreman' not in options.skip:
+        if 'puppet' not in options.skip:
```
evgeni, line 1478: @wzzrd or do you remember why you implemented it that way?!

ehelms (approved):

bootstrap.py
wzzrd: No, sorry. Don't remember.

PR #310, opened 2019-05-16
BZ #1710546 - decode ALL results we get from rpm

evgeni:
RPM in RHEL8 returns bytes, not strings, but we want to compare strings


evgeni:
meh, pylint

ehelms (approved):
pending tests

evgeni:
tests should be fixed when https://github.com/Katello/katello-client-bootstrap/pull/311 is merged

PR #312, opened 2019-06-07
Fix typo

ogajduse:
Would be good if some native speaker could check this.


sideangleside:
Looks good to me. merging. Thanks for the PR @ogajduse

PR #313, opened 2019-06-27
BZ#1717096: only pass --remove-rhn-packages when also doing legacy purge

evgeni:
rhn-migrate-classic-to-rhsm cannot be called with both --keep and
--remove-rhn-packages


sideangleside:
Seems duplicitous to do this twice. If I pass `--legacy-purge` the rhn-migrate command removes legacy packages, and then we _also_ attempt to remove them in the `remove_obsolete_packages()` function. 

That seems weird. But people also wear belts and suspenders. The code is fine though. No objections there.

sideangleside (approved):

evgeni:
The idea behind this was to benefit of a (possibly) updated package list in the migration tool, while still being able to clean up when we don't use it.

PR #319, opened 2019-09-09
Release 1.7.3

evgeni:
* BZ #1659414 - don't fail on unknown smart proxies (#315)
* BZ #1729179 - don't migrate EL8 hosts (#317)
* Puppet 6 support (#316)
* BZ#1717096: only pass --remove-rhn-packages when also doing legacy purge (#313)
* Open file in 'w' mode instead of 'wb' (#314)
* BZ #1710546 - decode ALL results we get from rpm (#310)
* update puppet config when puppet is not skipped, not foreman
* move enable_rhsmcertd just after doing the subscription stuff
* only execute API calls of "new capsule" mode when Foreman is not skipped
* improve error message when running on an registered host

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -63,7 +63,7 @@ except ImportError:
 import rpm  # pylint:disable=import-error
 
 
-VERSION = '1.7.2'
+VERSION = '1.7.3'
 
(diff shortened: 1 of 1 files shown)


sideangleside:
:+1:

ehelms (approved):

PR #283, opened 2018-12-19
[RFE] add a capability in the bootstrap script to 'preserve proxy' when reconfiguring/migrating the client.

wbclark:
This PR adds an option `--preserve-rhsm-proxy` which preserves the proxy settings in `/etc/rhsm/rhsm.conf`

This was originally requested in https://bugzilla.redhat.com/show_bug.cgi?id=1647631

PR #326, opened 2020-03-23
Fixes #325 - add persistent sub-man settings in bootstrap.py

### LINKED ISSUE #325: subscription-manager proxy entries are not persistent in rhsm.conf
Example:

 **Client** <----> **Proxy** <----> **Capsule**.

In this situation, passing in the RHSM proxy options, `--proxy=PROXY_URL` , `--proxyuser=PROXY_USER` and `--proxypassword=PROXY_PASSWORD`  are not persistent after the host is registered. 

Proxy settings are mandatory in this case as any subscription-manager requests must go through a proxy, as direct communication with the capsule isn't allowed.

snagoor:
Introduced new flags in bootstrap.py as below for persistent rhsm proxy settings

~~~
 --rhsm-proxy-name=RHSM_PROXY_NAME
                        Proxy Server Name which shall be passed to
                        subscription-manager
  --rhsm-proxy-port=RHSM_PROXY_PORT
                        Proxy Server Port which shall be passed to
                        subscription-manager
  --rhsm-proxy-user=RHSM_PROXY_USER
                        Proxy Server Username which shall be passed to
                        subscription-manager
  --rhsm-proxy-password=RHSM_PROXY_PASSWORD
                        Proxy Server Password which shall be passed to
                        subscription-manager
  --preserve-rhsm-proxy-settings
                        Preserve proxy settings in /etc/rhsm/rhsm.conf when
                        RHSM Proxy Settings are used

~~~

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -498,6 +498,20 @@ def register_systems(org_name, activationkey):
         options.smargs += " --release %s" % options.release
     if check_subman_version(SUBSCRIPTION_MANAGER_SERVER_TIMEOUT_VERSION):
         exec_failok("/usr/sbin/subscription-manager config --server.server_timeout=%s" % options.timeout)
+    if options.preserve_rhsm_proxy_settings:
+        exec_failok("/usr/sbin/subscription-manager config --server.proxy_hostname=%s" % options.rhsm_proxy_name)
+        exec_failok("/usr/sbin/subscription-manager config --server.proxy_port=%s" % options.rhsm_proxy_port)
+        exec_failok("/usr/sbin/subscription-manager config --server.proxy_user=%s" % options.rhsm_proxy_user)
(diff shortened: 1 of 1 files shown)


sideangleside:
Is this not what the `--preserve-rhsm-proxy` option does today?  (See https://github.com/Katello/katello-client-bootstrap/pull/283)

snagoor:
> Is this not what the `--preserve-rhsm-proxy` option does today? (See #283)

@sideangleside thank you for the review. You are correct https://github.com/Katello/katello-client-bootstrap/pull/283 does the job only when the user manually fills-in the proxy entries in rhsm.conf file and before executing the bootstrap.py script.

This is not ideal in a situation where an end user want to automate the registration of hosts, first they have to fill in all the proxy details in `rhsm.conf` on hosts and then execute the bootstrap script with `--preserve-rhsm-proxy` flag. So, having the necessary flags exposed as options would help them achieve both these things in one-go. 

Just want to improve the user experience here.

PR #327, opened 2020-03-24
Fixes #29396 - Add support for katello-host-tools-tracer

jturel:
Adds --install-katello-host-tools-tracer option:

```
[SUCCESS], [2020-03-24 00:36:16], [/usr/bin/yum -y install katello-host-tools-tracer], completed successfully.
```

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -532,6 +532,12 @@ def install_katello_host_tools():
     call_yum("install", "katello-host-tools")
 
 
+def install_katello_tracer():
+    """Install Katello Host Tools Tracer"""
+    print_generic("Installing Katello Host Tools Tracer")
+    call_yum("install", "katello-host-tools-tracer")
+
+
 def clean_puppet():
     """Remove old Puppet Agent and its configuration"""
     print_generic("Cleaning old Puppet Agent")
@@ -1245,6 +1251,7 @@ if __name__ == '__main__':
     parser.add_option("--ignore-registration-failures", dest="ignore_registration_failures", action="store_true", help="Continue running even if registration via subscription-manager/rhn-migrate-classic-to-rhsm returns a non-zero return code.")
     parser.add_option("--preserve-rhsm-proxy", dest="preserve_rhsm_proxy", action="store_true", help="Preserve proxy settings in /etc/rhsm/rhsm.conf when migrating RHSM -> RHSM")
     parser.add_option("--install-katello-agent", dest="install_katello_agent", action="store_true", help="Installs the Katello Agent", default=False)
+    parser.add_option("--install-katello-host-tools-tracer", dest="install_katello_host_tools_tracer", action="store_true", help="Installs the Katello Host Tools Tracer Plugin for Yum or DNF", default=False)
     (options, args) = parser.parse_args()
 
     if options.no_foreman:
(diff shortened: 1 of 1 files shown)


chris1984 (approved):
ACK, works good:

```bash
./bootstrap.py
  --install-katello-host-tools-tracer
                        Installs the Katello Host Tools Tracer Plugin for Yum
                        or DNF

./bootstrap.py -s sat-6-7-qa-rhel7.thinkstation.example.com -l admin -p changeme --content-only -o Default_Organization --install-katello-host-tools-tracer -a testkey

[SUCCESS], [2020-03-24 19:21:28], [/usr/bin/yum -y install katello-host-tools-tracer], completed successfully.

rpm -qa | grep katello-host-tools-tracer
katello-host-tools-tracer-3.5.1-2.el7sat.noarch
```

chris1984:
[test redmine]

jturel:
@evgeni @sideangleside any feedback for me here?

evgeni:
Can't this be solved via --install-packages? I'm hesitant adding another option :(

jturel:
> Can't this be solved via --install-packages? I'm hesitant adding another option :(

I'd like to advertise the utility - they probably don't know about it. We can remove --install-katello-agent soon so we're breaking even ;)

sideangleside:
Actually, I see no reason why we'd not install it by default, and provide a `--no-install-katello-host-tools-tracer` option to negate the install if needed by the admin. I think we should aim to have a pretty robust 'out of box' experience. Packages/Configuration related to managing the client well that do not affect its security posture should IMHO be default. 

So `katello-host-tools` and `katello-host-tools-tracer` are default. Deploying REX keys are opt-in. 



`--install-packages` was moreso intended to install additional non-Katello/Foreman packages, and while it _does_ work here, I almost think of it as optional.

jturel:
> Actually, I see no reason why we'd not install it by default, and provide a `--no-install-katello-host-tools-tracer` option to negate the install if needed by the admin. I think we should aim to have a pretty robust 'out of box' experience. Packages/Configuration related to managing the client well that do not affect its security posture should IMHO be default.
> 
> So `katello-host-tools` and `katello-host-tools-tracer` are default. Deploying REX keys are opt-in.
> 
> `--install-packages` was moreso intended to install additional non-Katello/Foreman packages, and while it _does_ work here, I almost think of it as optional.

Agree on the robust out of box experience. Check out SAT-E-83 for why I don't think we should enable this by default yet. Since the decision on SAT-E-83 I've been looking at how to improve the performance, but it appears that it's tied up in how tracer itself works. I'm still learning about it. Before enabling it by default I think it needs to be fast all the time, but that's not the case at the moment.

jturel:
@evgeni @sideangleside can we add this for now and post-next-release look at enabling tracer by default? Taking the feature out of tech preview in should see increased usage so we can address any underlying issues that have gone unnoticed. If all is well then I can definitely get behind enabling by default. Ping me if we need to hop on a call or something :)

chris1984:
ping @sideangleside @evgeni

evgeni:
I just don't think bootstrap is the right place for it. Bootstraps purpose is to make it easy to attach an existing system to Foreman/Katello, once this is done, you can use whatever method available (rex, ansible, puppet, chef, katello agent **shudder**) to install additional packages.

If tracer is supposed to be out of tech preview now, make it easy to deploy it from the UI, done :)

sideangleside:
I am OK with changing the defaults to deploy *katello-host-tools-tracer*. I am not OK with another CLI option for it. 

If there are concerns with the robustness or performance of katello-host-tools-tracer, then (IMHO) we put this PR on hold until those concerns are addressed.

jturel:
Thanks for the feedback. I do like the idea of making it easy to install via the UI. Closing this in favor of a solution along those lines.

PR #333, opened 2020-10-08
don't fail if sub-man-migration can't be installed

evgeni:


sbernhard:
LGTM. Better than the implementation before and it shouldn't break anything.

PR #337, opened 2020-11-04
properly load dnf repo substitutions from /etc

evgeni:
this is needed for Oracle Linux where the default repos have additional
substitutions which break package discovery

Replaces: #336


evgeni:
@sbernhard I think this should fix the issue you had without introducing any weird command line switches ;)

sbernhard:
Works. Thank you!

sbernhard (approved):

PR #348, opened 2021-09-06
don't fail decoding data from RPM

evgeni:
some versions of the RPM bindings will return strings, some bytes, some
unicode… And only some of these know decode().
So let's just catch that AttributeError and assume there is nothing to
decode if it's raised.

Fixes bootstrap on EL 8.5+ which got a "fixed" RPM.
See https://bugzilla.redhat.com/1840142 for some background.


evgeni:
CI fixes in #349

ehelms:
Needs rebase

evgeni:
> Needs rebase

it has been done

ehelms (approved):

PR #357, opened 2021-12-17
test on centos stream 8 and 9 instead of fedora and ubi

evgeni:


.github/workflows/main.yml
```diff
@@ -5,11 +5,22 @@ on: [push, pull_request]
...
+          - "lint"
+          - "test-centos:6"
+          - "test-centos:7"
+          - "test3-centos:stream8"
+          - "test3-centos:stream9"
     steps:
       - uses: actions/checkout@v2.4.0
       - name: Set up Python
         uses: actions/setup-python@v2.3.1
+        if: matrix.target == 'lint'
```
ekohl, line 21: Wouldn't it be cleaner to write a separate lint job for this? So
```yaml
jobs:
  lint:
    # ...

  test:
    # ...
```

ekohl (approved):

.github/workflows/main.yml
```diff
@@ -5,11 +5,22 @@ on: [push, pull_request]
...
+      fail-fast: false
+      matrix:
+        target:
+          - "lint"
+          - "test-centos:6"
+          - "test-centos:7"
+          - "test3-centos:stream8"
+          - "test3-centos:stream9"
     steps:
       - uses: actions/checkout@v2.4.0
```
ekohl, line 18: Oh, and unrelated but isn't this good enough:
```suggestion
      - uses: actions/checkout@v2
```

.github/workflows/main.yml
evgeni: It would, I was just avoiding duplication if the other steps

PR #362, opened 2022-03-16
detect the right api port, as it can differ from the rhsm port

evgeni:


evgeni:
this change might break compatibility with *really* old releases (I've tested it against Foreman 2.1/Katello 3.16), but I don't care ;-)

ehelms (approved):
Untested -- but I know you asked a user to try this out so feel free to merge if they do test and are successful

evgeni:
The user in https://community.theforeman.org/t/smart-proxy-missing-api-after-foreman-3-1-2-katello-4-3-upgrade/27717 reported success with this patch.

Note to myself before merging: add a comment about the currently known cases where the detection will fail.

PR #364, opened 2022-04-14
better handle setups without Puppet

evgeni:
- check for "Puppet" feature when moving between proxies
- don't setup Puppet agent when no Puppet environment could be found


swadeley (approved):
Works for me

ekohl (approved):
I didn't test this, but it reads ok.

evgeni:
FWIW, this is fixing https://bugzilla.redhat.com/show_bug.cgi?id=2075434

